PR #55, opened 2019-11-03
Enhanced/automatically tab to edit

mohammedhemaid:
A video for it
https://photos.app.goo.gl/223pjfBTEXGt3TPG9

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -42,6 +42,9 @@ import uk.co.deanwild.materialshowcaseview.MaterialShowcaseView;
 public class MainActivity extends StandardActivity {
 
     public static final String LIST_NAME_KEY = "listName";
+    public static final int NAME_CHOOSING_FRAGMENT = 0;
+    public static final int EDIT_NAME_LIST_FARGMENT = 1;
+    public static final String LIST_TYPE = "listType";
 
     @BindView(R.id.coordinator_layout) View parent;
     @BindView(R.id.focal_point) View focalPoint;
@@ -166,6 +169,7 @@ public class MainActivity extends StandardActivity {
         Intent intent = new Intent(this, ListActivity.class);
         String listName = nameListsAdapter.getItem(position);
         intent.putExtra(LIST_NAME_KEY, listName);
+        intent.putExtra(LIST_TYPE, NAME_CHOOSING_FRAGMENT);
         intent.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT);
         startActivity(intent);
     }
@@ -183,6 +187,7 @@ public class MainActivity extends StandardActivity {
             nameListsAdapter.addList(newList);
             Intent intent = new Intent(this, ListActivity.class);
             intent.putExtra(LIST_NAME_KEY, newList);
+            intent.putExtra(LIST_TYPE, EDIT_NAME_LIST_FARGMENT);
             startActivity(intent);
         }
     }
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -40,11 +40,11 @@ import uk.co.deanwild.materialshowcaseview.MaterialShowcaseSequence;
 import uk.co.deanwild.materialshowcaseview.MaterialShowcaseView;
 
 import static com.randomappsinc.studentpicker.listpage.ListActivity.LIST_TYPE;
+import static com.randomappsinc.studentpicker.listpage.ListActivity.START_ON_EDIT_PAGE;
 
 public class MainActivity extends StandardActivity {
 
     public static final String LIST_NAME_KEY = "listName";
-    public static final boolean START_ON_EDIT_PAGE = true;
 
     @BindView(R.id.coordinator_layout) View parent;
     @BindView(R.id.focal_point) View focalPoint;
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -176,7 +176,7 @@ public class MainActivity extends StandardActivity {
             nameListsAdapter.addList(newList);
             Intent intent = new Intent(this, ListActivity.class);
             intent.putExtra(LIST_NAME_KEY, newList);
-            intent.putExtra(START_ON_EDIT_PAGE , true );
+            intent.putExtra(START_ON_EDIT_PAGE, true);
             startActivity(intent);
         }
     }
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -39,7 +39,6 @@ import uk.co.deanwild.materialshowcaseview.IShowcaseListener;
 import uk.co.deanwild.materialshowcaseview.MaterialShowcaseSequence;
 import uk.co.deanwild.materialshowcaseview.MaterialShowcaseView;
 
-import static com.randomappsinc.studentpicker.listpage.ListActivity.LIST_TYPE;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -259,6 +264,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 return true;
             case R.id.view_tutorial:
                 showTutorial(false);
+                UIUtils.hideKeyboard(this);
```
Gear61: If you update this branch against Gear61:master, this should go away. It's not part of this pull request, so let's try to get it out to keep the pull request focused. At Facebook, we have a concept of "One and only thesis", which states that every pull request should be hyper-focused and only do 1 thing.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -42,6 +42,9 @@
 public class MainActivity extends StandardActivity {
 
     public static final String LIST_NAME_KEY = "listName";
+    public static final int NAME_CHOOSING_FRAGMENT = 0;
```
Gear61: Some thoughts here:
1. I feel like the receiver (ListActivity) should contain the keys, not the sender (MainActivity). Putting the keys within the sender makes this functionality less flexible in the long run. If ListActivity can be started from another place in the future, we will need to define a new set of keys, which will lead to dangerous copy-paste. Think about this problem as ListActivity "exposing" an API for other places to call to control its UI/UX.
2. As mentioned in class, each component should "hide" as much of its functionality as possible. By passing in the ViewPager index indirectly, ListActivity is revealing that it uses a ViewPager. This is fragile in the future since ListActivity may move off of a ViewPager implementation. Let's just create a new key like "START_ON_EDIT_PAGE" and the value will be a true/false boolean.
3. Do the simple thing first! Right now, there isn't a need to expose such a powerful API since by default, the ViewPager starts off on index 0 and we only need to make it 1 every once in a while. A powerful API also comes with risk, since a caller might pass in 2 or something and cause index out of bounds.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -166,6 +169,7 @@ public void onItemClick(int position) {
         Intent intent = new Intent(this, ListActivity.class);
         String listName = nameListsAdapter.getItem(position);
         intent.putExtra(LIST_NAME_KEY, listName);
+        intent.putExtra(LIST_TYPE, NAME_CHOOSING_FRAGMENT);
```
Gear61: Once we move to a simple, declarative boolean passing model, we won't need this line. 👍

Gear61 (changes requested):
This is a good start! Love the recorded video as the test plan (shows that your code works) and I see that you regression tested the other case as well. Hope you learn from the comments I left.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
mohammedhemaid: I think I did a mistake here, Should I merge the branch with the master and then continue the second issue?
Or make the last commit as ahead for the new branch and then continue from there

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -174,6 +177,7 @@ public void addItem() {
             nameListsAdapter.addList(newList);
             Intent intent = new Intent(this, ListActivity.class);
             intent.putExtra(LIST_NAME_KEY, newList);
+            intent.putExtra(LIST_TYPE, START_ON_EDIT_PAGE);
```
Gear61: Sorry for the miscommunication; I meant that the key should be:

`public static final String START_ON_EDIT_PAGE = "startOnEditPage"`

And we just pass in true as the value here.

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListActivity.java
```diff
@@ -42,12 +46,17 @@ protected void onCreate(Bundle savedInstanceState) {
         String listName = getIntent().getStringExtra(MainActivity.LIST_NAME_KEY);
         setTitle(listName);
 
+        boolean startFormEditPage = getIntent().getBooleanExtra(LIST_TYPE,false);
```
Gear61:
```suggestion
        boolean startFromEditPage = getIntent().getBooleanExtra(START_ON_EDIT_PAGE, false);
```

Gear61 (changes requested):
Thanks for incorporating the feedback! Should be good after 1 more round of changes.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -174,6 +176,7 @@ public void addItem() {
             nameListsAdapter.addList(newList);
             Intent intent = new Intent(this, ListActivity.class);
             intent.putExtra(LIST_NAME_KEY, newList);
+            intent.putExtra(START_ON_EDIT_PAGE , true );
```
Gear61:
```suggestion
            intent.putExtra(START_ON_EDIT_PAGE, true);
```

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListActivity.java
```diff
@@ -42,12 +45,17 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
+        boolean startFormEditPage = getIntent().getBooleanExtra(START_ON_EDIT_PAGE,false);
+
         preferencesManager = new PreferencesManager(this);
         ListTabsAdapter listTabsAdapter = new ListTabsAdapter(
                 getSupportFragmentManager(),
                 listName,
                 listTabTitles);
         nameListPager.setAdapter(listTabsAdapter);
+        if (startFormEditPage){
```
Gear61: nit: I think we should just inline getIntent().getBooleanExtra(START_ON_EDIT_PAGE, false) within this if statement.

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListActivity.java
```diff
@@ -21,6 +21,9 @@
 
 public class ListActivity extends StandardActivity implements ShakeDetector.Listener {
 
+    public static final int EDIT_NAME_LIST_FRAGMENT = 1;
```
Gear61: nit: Removing fragment from the name hides implementation, so I think this is slightly cleaner.
```suggestion
    public static final int EDIT_NAME_LIST_PAGE_POSITION = 1;
```

Gear61 (changes requested):
Looks a lot better! Just a couple small changes left.

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListActivity.java
```diff
@@ -21,6 +21,9 @@
 
 public class ListActivity extends StandardActivity implements ShakeDetector.Listener {
 
+    public static final int EDIT_NAME_LIST_PAGE_POSITION = 1;
```
Gear61: nit: This should be private. Let's hide the implementation.
```suggestion
    private static final int EDIT_NAME_LIST_PAGE_POSITION = 1;
```

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListActivity.java
```diff
@@ -48,6 +51,9 @@ protected void onCreate(Bundle savedInstanceState) {
                 listName,
                 listTabTitles);
         nameListPager.setAdapter(listTabsAdapter);
+        if (getIntent().getBooleanExtra(START_ON_EDIT_PAGE,false)){
```
Gear61: nit: Small spacing issue
```suggestion
        if (getIntent().getBooleanExtra(START_ON_EDIT_PAGE, false)) {
```

Gear61 (changes requested):
Please increase the versioning as well. Should be versionName of "3.9.11" and versionCode of 401.

Gear61 (approved):
Awesome! 👍

PR #60, opened 2019-11-08
Refactor Base Adapter to Recycle View and animate the changes.

mohammedhemaid:
sorry for these "Clean the code and use boolean" and "merge branch" commits, I forget to merge them in my master, So when I was syncing the fork some conflict shows up so I had to do that.

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -39,6 +39,8 @@ import uk.co.deanwild.materialshowcaseview.IShowcaseListener;
 import uk.co.deanwild.materialshowcaseview.MaterialShowcaseSequence;
 import uk.co.deanwild.materialshowcaseview.MaterialShowcaseView;
 
+import static com.randomappsinc.studentpicker.listpage.ListActivity.LIST_TYPE;
+
 public class MainActivity extends StandardActivity {
 
     public static final String LIST_NAME_KEY = "listName";
@@ -174,6 +176,7 @@ public class MainActivity extends StandardActivity {
             nameListsAdapter.addList(newList);
             Intent intent = new Intent(this, ListActivity.class);
             intent.putExtra(LIST_NAME_KEY, newList);
+            intent.putExtra(LIST_TYPE,ListActivity.START_ON_EDIT_PAGE);
             startActivity(intent);
         }
     }
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -10,11 +10,11 @@ import android.view.MenuItem;
 import android.view.View;
 import android.widget.EditText;
 import android.widget.ImageView;
-import android.widget.ListView;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.core.app.ActivityCompat;
+import androidx.recyclerview.widget.RecyclerView;
 
 import com.afollestad.materialdialogs.MaterialDialog;
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
@@ -34,21 +34,20 @@ import butterknife.BindString;
 import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
-import butterknife.OnItemClick;
 import uk.co.deanwild.materialshowcaseview.IShowcaseListener;
 import uk.co.deanwild.materialshowcaseview.MaterialShowcaseSequence;
 import uk.co.deanwild.materialshowcaseview.MaterialShowcaseView;
 
 import static com.randomappsinc.studentpicker.listpage.ListActivity.START_ON_EDIT_PAGE;
 
-public class MainActivity extends StandardActivity {
+public class MainActivity extends StandardActivity implements NameListsAdapter.OnListItemClickListener {
 
     public static final String LIST_NAME_KEY = "listName";
 
     @BindView(R.id.coordinator_layout) View parent;
     @BindView(R.id.focal_point) View focalPoint;
     @BindView(R.id.item_name_input) EditText newListInput;
-    @BindView(R.id.content_list) ListView lists;
+    @BindView(R.id.names_recycle_view) RecyclerView recyclerView;
     @BindView(R.id.no_content) TextView noContent;
     @BindView(R.id.add_item) View addItem;
     @BindView(R.id.plus_icon) ImageView plus;
@@ -73,8 +72,8 @@ public class MainActivity extends StandardActivity {
                 this,
                 IoniconsIcons.ion_android_upload).colorRes(R.color.white));
 
-        nameListsAdapter = new NameListsAdapter(this, noContent);
-        lists.setAdapter(nameListsAdapter);
+        nameListsAdapter = new NameListsAdapter(this,this, noContent);
+        recyclerView.setAdapter(nameListsAdapter);
 
         if (preferencesManager.getFirstTimeUser()) {
             preferencesManager.setFirstTimeUser(false);
@@ -109,7 +108,8 @@ public class MainActivity extends StandardActivity {
                 .setContentText(R.string.import_explanation)
                 .setListener(new IShowcaseListener() {
                     @Override
-                    public void onShowcaseDisplayed(MaterialShowcaseView materialShowcaseView) {}
+                    public void onShowcaseDisplayed(MaterialShowcaseView materialShowcaseView) {
+                    }
 
                     @Override
                     public void onShowcaseDismissed(MaterialShowcaseView materialShowcaseView) {
@@ -137,7 +137,7 @@ public class MainActivity extends StandardActivity {
                 .negativeText(R.string.no_im_good)
                 .positiveText(R.string.will_rate)
                 .onPositive((dialog, which) -> {
-                    Uri uri =  Uri.parse("market://details?id=" + getApplicationContext().getPackageName());
+                    Uri uri = Uri.parse("market://details?id=" + getApplicationContext().getPackageName());
                     Intent intent = new Intent(Intent.ACTION_VIEW, uri);
                     if (!(getPackageManager().queryIntentActivities(intent, 0).size() > 0)) {
                         UIUtils.showSnackbar(parent, getString(R.string.play_store_error));
@@ -154,7 +154,7 @@ public class MainActivity extends StandardActivity {
         nameListsAdapter.refreshList();
     }
 
-    @OnItemClick(R.id.content_list)
+    @Override
     public void onItemClick(int position) {
         Intent intent = new Intent(this, ListActivity.class);
         String listName = nameListsAdapter.getItem(position);
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -108,8 +108,7 @@ public class MainActivity extends StandardActivity implements NameListsAdapter.O
                 .setContentText(R.string.import_explanation)
                 .setListener(new IShowcaseListener() {
                     @Override
-                    public void onShowcaseDisplayed(MaterialShowcaseView materialShowcaseView) {
-                    }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -137,7 +136,7 @@ public void showPleaseRateDialog() {
                 .negativeText(R.string.no_im_good)
                 .positiveText(R.string.will_rate)
                 .onPositive((dialog, which) -> {
-                    Uri uri =  Uri.parse("market://details?id=" + getApplicationContext().getPackageName());
+                    Uri uri = Uri.parse("market://details?id=" + getApplicationContext().getPackageName());
```
Gear61: Thanks for making this change! So before, I think we had to use getApplicationContext() because the code was nested within an anonymous class. Now that it's more exposed with lambda, I think we can do getPackageName() directly since we're within an activity. Try this and if it compiles, let's make the change.

```suggestion
                    Uri uri = Uri.parse("market://details?id=" + getPackageName());
```

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -40,6 +42,24 @@ public NameListsAdapter(Context context, TextView noContent) {
...
+    @NonNull
+    @Override
+    public NameListsAdapter.NameListViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+        View view = LayoutInflater.from(parent.getContext()).inflate(R.layout.names_list_item, parent, false);
+        return new NameListViewHolder(view);
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull NameListsAdapter.NameListViewHolder holder, int position) {
+
```
Gear61: nit: Delete this newline?

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -117,19 +126,24 @@ public void onClick(@NonNull MaterialDialog dialog, @NonNull DialogAction which)
                         preferencesManager.removeNameList(getItem(position));
                         content.remove(position);
                         setNoContent();
-                        notifyDataSetChanged();
+                        notifyItemRemoved(position);
+                        notifyItemRangeChanged(position, content.size());
```
Gear61: I don't think we need the notifyItemRangeChanged. We're only removing 1 item right? We aren't updating several.

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -117,19 +126,24 @@ public void onClick(@NonNull MaterialDialog dialog, @NonNull DialogAction which)
...
+
```
Gear61: nit: Also, delete this newline?

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -146,20 +160,19 @@ public void renameList() {
...
+                if (position != RecyclerView.NO_POSITION) {
+                    listItemClickListener.onItemClick(position);
+                }
+            }
         }
-        holder.loadList(position);
-        return view;
+    }
+
+    public interface OnListItemClickListener {
```
Gear61: public interfaces should go at the top of the class, even above the member variables. The reason for this is that public interfaces are very important (they define the contract between 2+ classes), so they should be at the top for easy access.

app/src/main/res/layout/activity_main.xml
```diff
@@ -1,12 +1,85 @@
...
+                >
+
+                <ImageView
+                    android:id="@+id/plus_icon"
+                    android:layout_width="match_parent"
+                    android:layout_height="match_parent"
+                    />
+            </androidx.cardview.widget.CardView>
+
+            <AutoCompleteTextView
```
Gear61: Looks like you copy-pasted from the list editing layout, which is a good start (as I mentioned before, it's very important to keep the code consistent), but we don't need this AutoCompleteTextView since it's for name suggestion on the list editing page (in case you want to enter in multiple copies of the same name for a raffle or something).

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -146,20 +160,19 @@ public void renameList() {
...
-        if (view == null) {
-            LayoutInflater vi = (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
-            view = vi.inflate(R.layout.names_list_item, parent, false);
-            holder = new NameListViewHolder(view);
-            view.setTag(holder);
-        } else {
-            holder = (NameListViewHolder) view.getTag();
+        @Override
+        public void onClick(View v) {
+            if (listItemClickListener != null) {
```
Gear61: This should never be null; let's not check against that. Nesting makes the code less readable, so we have nothing to gain here and something to lose.

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -146,20 +160,19 @@ public void renameList() {
...
+                int position = getAdapterPosition();
+                if (position != RecyclerView.NO_POSITION) {
```
Gear61: I have also personally never seen this case happen; we should remove this check. When it comes to bugs, we should fail fast. This means that if there's some weird case that should never happen, we shouldn't guard against it with an if check; we should just let it happen and crash the app. This allows us to find bugs faster instead of suppressing them and most likely leaving the user in a weird UI state which they may not report.

app/src/main/res/layout/activity_main.xml
```diff
@@ -1,12 +1,85 @@
...
+                />
+        </FrameLayout>
+
+        <TextView
+            android:id="@+id/no_content"
+            style="@style/empty_state_text"
+            />
+
+        <TextView
+            android:id="@+id/num_names"
```
Gear61: This also isn't used on the homepage. Let's remove it.

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -23,15 +23,17 @@
 import butterknife.ButterKnife;
 import butterknife.OnClick;
 
-public class NameListsAdapter extends BaseAdapter {
+public class NameListsAdapter extends RecyclerView.Adapter<NameListsAdapter.NameListViewHolder> {
 
     private Context context;
```
Gear61, line 32: Having adapters store a reference to a context is a legacy way of doing things when I frankly didn't know any better, haha. It's dangerous to store contexts because they are often times activities and if you don't clean up references to them properly, you leak an entire activity, which is massive (you saw me do this in TravelBuddy when you ran LeakCanary on it, haha). Given that we're doing a refactor, let's take the opportunity to leave the code in a better place than which we found it and remove this reference and the parameter it adds in the constructor.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -73,8 +72,8 @@ protected void onCreate(Bundle savedInstanceState) {
                 this,
                 IoniconsIcons.ion_android_upload).colorRes(R.color.white));
 
-        nameListsAdapter = new NameListsAdapter(this, noContent);
-        lists.setAdapter(nameListsAdapter);
+        nameListsAdapter = new NameListsAdapter(this,this, noContent);
```
Gear61: nit: We're going to end up changing this but leaving a reminder to space things like this when we get here.

```suggestion
        nameListsAdapter = new NameListsAdapter(this, this, noContent);
```

Gear61 (changes requested):
I know I left a lot of comments, but this is an awesome start! This wasn't an easy change to make, but you got it working! This is a very important change for this app to take the codebase into the present and improve the user experience, so I'm really glad we're working on this.

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
mohammedhemaid: when I used the notifyItemRemoved the items still in their place so I have to rearrange them.

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
mohammedhemaid: The context was Already there haha, you used it to pass it to the dialogs, I was thinking to make a reference to the parent context to use the dialog inside the Adapter or move them to the MainActivity and add abstract methods to the listener.

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
Gear61: Interesting... I've written this code before and I don't remember having to do that. I wonder if it's an Android 10 issue or something wrong with my old, janky code lol.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -66,15 +65,14 @@ protected void onCreate(Bundle savedInstanceState) {
         ButterKnife.bind(this);
 
         preferencesManager = new PreferencesManager(this);
-        newListInput.setHint(R.string.add_list_hint);
         noContent.setText(R.string.no_lists_message);
```
Gear61: Since the noContent view isn't shared anymore, let's set this in XML instead.

app/src/main/res/layout/activity_main.xml
```diff
@@ -1,12 +1,79 @@
...
+                />
+        </FrameLayout>
+
+        <TextView
+            android:id="@+id/no_content"
+            style="@style/empty_state_text"
+            />
+
+        <androidx.recyclerview.widget.RecyclerView
+            android:id="@+id/names_recycle_view"
```
Gear61: nit: I'm not a fan of naming things after the type of view; it doesn't convey new information since we will bind this to a RecyclerView. I think a name like "user_lists" is better; it says what this view contains (the lists of the user).

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -117,19 +129,23 @@ public void onClick(@NonNull MaterialDialog dialog, @NonNull DialogAction which)
                         preferencesManager.removeNameList(getItem(position));
                         content.remove(position);
                         setNoContent();
-                        notifyDataSetChanged();
+                        notifyItemRemoved(position);
+                        notifyItemRangeChanged(position, 1);
```
Gear61, line 133: What kind of bug happened if we didn't have this line again? The UI wouldn't update with the removed row?

Gear61 (changes requested):
Almost there, just a couple more changes and we should be good to go!

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
mohammedhemaid: Yes, the position will not change I will make a video of the bug.

mohammedhemaid:
this is the error when I remove the notifyItemRangeChanged(position, 1);


    Process: com.randomappsinc.studentpicker, PID: 16914
    java.lang.IndexOutOfBoundsException: Index: 2, Size: 2
        at java.util.ArrayList.get(ArrayList.java:437)
        at com.randomappsinc.studentpicker.home.NameListsAdapter.getItem(NameListsAdapter.java:89)
        at com.randomappsinc.studentpicker.home.NameListsAdapter$3.onClick(NameListsAdapter.java:129)
        at com.afollestad.materialdialogs.MaterialDialog.onClick(MaterialDialog.java:410)

mohammedhemaid:
this is a video showing the crash:
https://photos.app.goo.gl/9AMNZPfd6cRxiHwx7

Gear61 (approved):
Thanks for doing this! :)

PR #62, opened 2019-11-28
tap voice entry for the name creation feature

mohammedhemaid:
The Video of the feature:
https://photos.app.goo.gl/mu5uZKQW3fgRF9KL7

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -1,16 +1,19 @@
 package com.randomappsinc.studentpicker.home;
 
 import android.Manifest;
+import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.os.Bundle;
+import android.speech.RecognizerIntent;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
 import android.widget.EditText;
 import android.widget.ImageView;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.core.app.ActivityCompat;
@@ -31,6 +34,9 @@ import com.randomappsinc.studentpicker.utils.PreferencesManager;
 import com.randomappsinc.studentpicker.utils.UIUtils;
 import com.randomappsinc.studentpicker.views.SimpleDividerItemDecoration;
 
+import java.util.List;
+import java.util.Locale;
+
 import butterknife.BindString;
 import butterknife.BindView;
 import butterknife.ButterKnife;
@@ -44,6 +50,8 @@ import static com.randomappsinc.studentpicker.listpage.ListActivity.START_ON_EDI
 public class MainActivity extends StandardActivity implements NameListsAdapter.Delegate {
 
     public static final String LIST_NAME_KEY = "listName";
+    private static final int SPEECH_REQUEST_CODE = 1;
+    private static final int IMPORT_FILE_REQUEST_CODE = 2;
 
     @BindView(R.id.coordinator_layout) View parent;
     @BindView(R.id.focal_point) View focalPoint;
@@ -194,6 +202,11 @@ public class MainActivity extends StandardActivity implements NameListsAdapter.D
         }
     }
 
+    @OnClick(R.id.voice_entry_icon)
+    public void voiceEntry() {
+        showGoogleSpeechDialog();
+    }
+
     @OnClick(R.id.import_text_file)
     public void importTextFile() {
         if (PermissionUtils.isPermissionGranted(Manifest.permission.READ_EXTERNAL_STORAGE, this)) {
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -61,4 +62,12 @@
 
     <string name="rename_list">Rename List</string>
     <string name="delete_list">Delete List</string>
+
+    <string name="speech_message">Tell us what kind of place you\'re looking for.</string>
```
Gear61: Let's customize the copy-pasted code to match our use case. :)
```suggestion
    <string name="list_name_speech_prompt_message">Tell us what you want to name your list</string>
```

app/src/main/res/layout/activity_main.xml
```diff
@@ -60,6 +60,17 @@
...
+
+            <com.joanzapata.iconify.widget.IconTextView
+                android:id="@+id/voice_entry_icon"
+                android:text="@string/voice_icon"
+                android:layout_gravity="end|center_vertical"
+                style="@style/list_item_button"
+                android:layout_marginEnd="60dp"
+                android:layout_marginRight="60dp"
+                android:textSize="24sp"
+                android:foreground="?attr/selectableItemBackgroundBorderless"
```
Gear61: Why selectableItemBackgroundBorderless instead of selectableItemBackground?

app/src/main/res/layout/activity_main.xml
```diff
@@ -60,6 +60,17 @@
                 android:layout_height="44dp"
                 android:hint="@string/add_list_hint"
                 />
+
+            <com.joanzapata.iconify.widget.IconTextView
+                android:id="@+id/voice_entry_icon"
+                android:text="@string/voice_icon"
+                android:layout_gravity="end|center_vertical"
+                style="@style/list_item_button"
```
Gear61: I don't think we should use this style. We're not within a list and we don't need the centerVertical attribute within it (that's a RelativeLayout only attribute). Let's just inline those attributes.

app/src/main/res/layout/activity_main.xml
```diff
@@ -60,6 +60,17 @@
...
+                android:layout_marginEnd="60dp"
```
Gear61: This should be 55dp I think so there isn't a gap between the icon space and the add button.

app/src/main/res/layout/activity_main.xml
```diff
@@ -60,6 +60,17 @@
                 android:layout_height="44dp"
                 android:hint="@string/add_list_hint"
                 />
+
+            <com.joanzapata.iconify.widget.IconTextView
+                android:id="@+id/voice_entry_icon"
+                android:text="@string/voice_icon"
```
Gear61, line 71: Let's make the textColor blue so it's more obvious that it's a button.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -229,19 +242,54 @@ public void onRequestPermissionsResult(
...
-            if (!filePath.endsWith(".txt")) {
-                UIUtils.showSnackbar(parent, getString(R.string.invalid_file));
-            } else {
-                Intent intent = new Intent(this, ImportFileActivity.class);
-                intent.putExtra(ImportFileActivity.FILE_PATH_KEY, filePath);
-                startActivity(intent);
-            }
+
+        switch (requestCode) {
+
```
Gear61: nit: Delete this newline.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -229,19 +242,54 @@ public void onRequestPermissionsResult(
...
+                        UIUtils.showSnackbar(parent, getString(R.string.invalid_file));
+                    } else {
+                        Intent intent = new Intent(this, ImportFileActivity.class);
+                        intent.putExtra(ImportFileActivity.FILE_PATH_KEY, filePath);
+                        startActivity(intent);
+                    }
+                }
+                break;
         }
+
```
Gear61: nit: Delete this newline.

Gear61 (changes requested):
I think we'll need to increase the marginRight on the EditText as well (I can't comment on lines not in the PR for some reason) so if the user types a lot, the text doesn't overlap with the new speech input button (as I mentioned in class, we need to always think about the edge cases!). I think we'll need to increase it to 99dp (55dp + 44dp). Just make sure to test the case to make sure it works.

app/src/main/res/layout/activity_main.xml
mohammedhemaid: yes, there is no need for selectableItemBackgroundBorderless, I just wanted the animation to be circular but it didn't work

app/src/main/res/layout/activity_main.xml
mohammedhemaid: image

image
this is the difference between the center and without the center or centerVertical

Gear61: Yeah, we still need the layout_gravity of center_vertical|end (for FrameLayouts), but we don't need the one within the style (for RelativeLayouts).

mohammedhemaid: Got it, I saw it but I think it will not make a difference so I kept it, but putting them inline it will be better and it will reduce the unused code.

mohammedhemaid:
I also cannot comment on the editText!
I changed it like this: 

image

The video of the clicking:
https://photos.app.goo.gl/MY3eEKxcutoe3akz5

app/src/main/res/layout/activity_main.xml
```diff
@@ -53,13 +53,27 @@
                 android:textColor="@color/dark_gray"
                 android:textSize="18dp"
                 android:background="@drawable/add_item_border"
-                android:layout_marginRight="55dp"
+                android:layout_marginRight="97dp"
```
Gear61: Thank you so much for the video! That is not the look I was going for; I suggested the wrong code, sorry! :(

We should keep the 55dp marginRight so the input doesn't overlap with the button but add a paddingRight of 44dp as well so the typed in text doesn't overlap with the microphone icon (the icon is still within the text input).

app/src/main/res/values/strings.xml
```diff
@@ -61,4 +62,12 @@
 
     <string name="rename_list">Rename List</string>
     <string name="delete_list">Delete List</string>
+
+    <string name="speech_message">Tell us what you want to name your list</string>
```
Gear61: speech_message is too generic, and we want to add speech input to other parts of the app.
```suggestion
    <string name="list_name_input_speech_message">Tell us what you want to name your list</string>
```

Gear61 (changes requested):
Thanks for addressing the feedback! Just a couple more changes, and we should be good to go.

app/src/main/res/layout/activity_main.xml
mohammedhemaid: The video of the update:
https://photos.app.goo.gl/wVeuWHSdMDKFrtNP8

Gear61 (approved):
Looks good! Merging this in now. :)

PR #66, opened 2019-12-05
Adding TextToSpeechManager to PresentationActivity.

alaaalzibda:
After the DRUMROLL player is completed I call the speak function which is inside the textToSpeechManager

### DIFF
--- a/Random-Name-Picker.iml
+++ b/Random-Name-Picker.iml
@@ -12,6 +12,7 @@
     <exclude-output />
     <content url="file://$MODULE_DIR$">
       <excludeFolder url="file://$MODULE_DIR$/.gradle" />
+      <excludeFolder url="file://$MODULE_DIR$/build" />
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
--- a/app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingFragment.java
@@ -2,6 +2,7 @@ package com.randomappsinc.studentpicker.choosing;
 
 import android.content.Intent;
 import android.os.Bundle;
+import android.util.Log;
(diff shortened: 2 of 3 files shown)


Random-Name-Picker.iml
```diff
@@ -12,6 +12,7 @@
     <exclude-output />
     <content url="file://$MODULE_DIR$">
       <excludeFolder url="file://$MODULE_DIR$/.gradle" />
+      <excludeFolder url="file://$MODULE_DIR$/build" />
```
Gear61, line 15: Hmmm, I need to remove this from the repository. It's an IDE specific file.

app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingFragment.java
```diff
@@ -2,6 +2,7 @@
 
 import android.content.Intent;
 import android.os.Bundle;
+import android.util.Log;
```
Gear61, line 5: Please delete this, it's not being used anywhere.

app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingFragment.java
```diff
@@ -148,7 +149,6 @@ public void choose() {
         if (nameChoosingAdapter.getCount() == 0) {
             return;
         }
-
```
Gear61, line 151: nit: Please put this new line back?

Gear61 (changes requested):
This is a good start, but we need to remove the old implementation of this app speaking feature. You need to remove the TextToSpeech instance from this activity and slot in to where it was being invoked before.

Random-Name-Picker.iml
alaaalzibda: Oh, I keep forgetting it, even in the new commit I forgot it, sorry
Do you add /Random-Name-Picker.iml inside the .gitignore file and then you call 
git add . as usual or git add . then git reset -- Random-name-picker.iml ?

PR #67, opened 2019-12-09
Adding textToSeechManager and removing TextToSpeech instance

alaaalzibda:


alaaalzibda:
Forget about the commits about lines, removing and returning the UIUtils
I was just trying something on git!

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
```diff
@@ -42,14 +43,17 @@
...
-        implements TextToSpeech.OnInitListener, ColorChooserDialog.ColorCallback {
+        implements ColorChooserDialog.ColorCallback, TextToSpeechManager.Listener {
 
     public static final String LIST_NAME_KEY = "listName";
     public static final String DRUMROLL_FILE_NAME = "drumroll.mp3";
 
-    @BindView(R.id.header) TextView header;
-    @BindView(R.id.names) TextView names;
-    @BindColor(R.color.dark_gray) int darkGray;
+    @BindView(R.id.header)
```
Gear61: nit: Can you undo these formatting changes? It makes it hard to figure out which bind corresponds to which variable.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
```diff
@@ -125,7 +125,6 @@ public void choose() {
 
     private void chooseNames() {
         names.clearAnimation();
-        textToSpeech.stop();
```
Gear61, line 128: We need to ask the TextToSpeechManager to stop speaking here.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
```diff
@@ -152,6 +151,7 @@ private void playSound() {
                     fileDescriptor.getLength());
             player.prepare();
             player.start();
+            player.setOnCompletionListener(mediaPlayer -> textToSpeechManager.speak(chosenNamesText));
```
Gear61: We shouldn't have this. The app will speak in the animation end block.

Gear61 (changes requested):
This is much better; thanks for making these changes! Normally I don't really care about commit history, but I feel like having 15 commits is overkill. After you make the changes for this code review, can you squash all your commits into 1? It should be something like this:

git reset --soft HEAD~15
git commit -am 'Refactor PresentationActivity to use TextToSpeechManager instead'
git push origin master --force

This should turn all your commits into 1 tidy commit with a great message. :) Make sure that 'origin' is the remote for your fork, otherwise you will overwrite history on the main repo.

Gear61 (changes requested):
There's a lot of changes in your pull request that are already in the core repo. I looked into your fork and the reason for this is that the master branch of your repo is behind that of the core repo. With any big project with a lot of collaborators, you need to make sure that your version of the code is always up to date with the shared version; otherwise your merge requests will be off.

This is tricky since your commits have extra information, so to save your changes, I would move your local copy of PresentationActivity somewhere else (otherwise you can write all your changes again after fixing your master branch). I assume your Git repo on your machine has 2 remotes, one for your fork (let's say it's named "origin") and one for the shared repo (let's say it's named "shared"). To update your master branch against that of the shared one, you would do this.

`git merge remotes/shared/master`

After you do that, you can put your changes to refactor PresentationActivity on top. Sorry for requesting all these changes, but I promise you that this working knowledge will be useful in the long run (this is how big teams in any company collaborate).

PR #69, opened 2019-12-12
[Refactor] Extract out edit and delete dialogs from the homepage adapter

mohammedhemaid:
this is a video shows that it works:
https://photos.app.goo.gl/MLSSC72F3m4ZEGsS8

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -56,6 +56,4 @@ dependencies {
     implementation 'com.nbsp:library:1.02'
     implementation 'com.github.deano2390:MaterialShowcaseView:1.2.0'
     implementation 'com.squareup:seismic:1.0.2'
-    //debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.0'
-
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -56,4 +56,6 @@ dependencies {
     implementation 'com.nbsp:library:1.02'
     implementation 'com.github.deano2390:MaterialShowcaseView:1.2.0'
     implementation 'com.squareup:seismic:1.0.2'
+    //debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.0'
+
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -7,8 +7,8 @@ android {
         applicationId "com.randomappsinc.studentpicker"
         minSdkVersion 16
         targetSdkVersion 28
-        versionCode 403
-        versionName "3.9.13"
+        versionCode 404
+        versionName "3.9.14"
     }
     buildTypes {
         release {
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
@@ -1,46 +1,37 @@
 package com.randomappsinc.studentpicker.home;
 
-import android.app.Dialog;
-import android.os.Bundle;
-
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.fragment.app.DialogFragment;
-import androidx.fragment.app.FragmentManager;
-
 import com.afollestad.materialdialogs.MaterialDialog;
 import com.randomappsinc.studentpicker.R;
 
-public class DeleteListDialog extends DialogFragment {
+public class DeleteListDialog {
 
     public interface Listener {
         void onDeleteListConfirmed(int position);
     }
 
-    private FragmentManager fragmentManager;
+    private MaterialDialog.Builder materialDialogBuilder;
     private Listener listener;
     private int position;
 
-    public DeleteListDialog(FragmentManager fragmentManager, Listener listener) {
-        this.fragmentManager = fragmentManager;
+    DeleteListDialog(MaterialDialog.Builder materialDialogBuilder, Listener listener) {
+        this.materialDialogBuilder = materialDialogBuilder;
         this.listener = listener;
+        onCreateDialog();
     }
 
-    @NonNull
-    @Override
-    public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
-        return new MaterialDialog.Builder(getActivity())
+   private void onCreateDialog() {
+        materialDialogBuilder
                 .title(R.string.confirm_deletion_title)
                 .content(R.string.confirm_deletion_message)
                 .positiveText(android.R.string.yes)
                 .negativeText(android.R.string.cancel)
                 .onPositive((dialog, which) ->
-                    listener.onDeleteListConfirmed(position)
+                        listener.onDeleteListConfirmed(position)
                 ).build();
     }
 
     public void show(int position) {
         this.position = position;
-        this.show(fragmentManager, null);
+        materialDialogBuilder.show();
     }
 }
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
@@ -1,0 +1,46 @@
+package com.randomappsinc.studentpicker.home;
+
+import android.app.Dialog;
+import android.os.Bundle;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.fragment.app.DialogFragment;
+import androidx.fragment.app.FragmentManager;
+
+import com.afollestad.materialdialogs.MaterialDialog;
+import com.randomappsinc.studentpicker.R;
+
+public class DeleteListDialog extends DialogFragment {
+
+    public interface Listener {
+        void onDeleteListConfirmed(int position);
+    }
+
+    private FragmentManager fragmentManager;
+    private Listener listener;
+    private int position;
+
+    public DeleteListDialog(FragmentManager fragmentManager, Listener listener) {
+        this.fragmentManager = fragmentManager;
+        this.listener = listener;
+    }
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
+        return new MaterialDialog.Builder(getActivity())
+                .title(R.string.confirm_deletion_title)
+                .content(R.string.confirm_deletion_message)
+                .positiveText(android.R.string.yes)
+                .negativeText(android.R.string.cancel)
+                .onPositive((dialog, which) ->
+                    listener.onDeleteListConfirmed(position)
+                ).build();
+    }
+
+    public void show(int position) {
+        this.position = position;
+        this.show(fragmentManager, null);
+    }
+}
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
@@ -181,12 +181,12 @@ public class MainActivity extends StandardActivity
 
     @Override
     public void onItemEditClick(int position) {
-        showRenameDialog(position);
+        renameListDialog.show(position);
     }
 
     @Override
     public void onItemDeleteClick(int position) {
-        showDeleteDialog(position);
+        deleteListDialog.show(position);
     }
 
     @Override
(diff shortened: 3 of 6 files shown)


app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -310,4 +323,42 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                    dialog.getActionButton(DialogAction.POSITIVE).setEnabled(submitEnabled);
+                })
+                .alwaysCallInputCallback()
+                .negativeText(android.R.string.cancel)
+                .onAny((dialog, which) -> {
+                    if (which == DialogAction.POSITIVE) {
+                        String newListName = dialog.getInputEditText().getText().toString().trim();
+                        dataSource.renameList(nameListsAdapter.getItem(position), newListName);
+                        preferencesManager.renameList(nameListsAdapter.getItem(position), newListName);
+                        nameListsAdapter.getContent().set(position, newListName);
```
Gear61: nit: Let's expose a method in the adapter called "renameItem(int position, String newName)". We are centralizing too much logic in the activity.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -310,4 +323,42 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+    private void showDeleteDialog(final int position) {
+        new MaterialDialog.Builder(this)
+                .title(R.string.confirm_deletion_title)
+                .content(R.string.confirm_deletion_message)
+                .positiveText(android.R.string.yes)
+                .negativeText(android.R.string.cancel)
+                .onPositive((dialog, which) -> {
+                    dataSource.deleteList(nameListsAdapter.getItem(position));
+                    preferencesManager.removeNameList(nameListsAdapter.getItem(position));
+                    nameListsAdapter.getContent().remove(position);
```
Gear61: nit: Same as above, let's expose a method in the adapter called "deleteItem(int position)".

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -310,4 +323,42 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
         return super.onOptionsItemSelected(item);
     }
+
+    private void showRenameDialog(final int position) {
+        new MaterialDialog.Builder(this)
```
Gear61: 364 lines is pretty large. Let's create a new class in the home package called "RenameListDialog". It will communicate with the activity through an interface (the method can be something like onRenameListConfirmed(int position, String newName). The activity will invoke the dialog through a method like show(int position). This new class will be responsible for creating and showing the dialog, allowing the activity to have less code. We also don't need to keep creating the same dialog multiple times. We can store a reference to the dialog in this new class and clear out the input whenever we need to show it again.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -310,4 +323,42 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                        dataSource.renameList(nameListsAdapter.getItem(position), newListName);
+                        preferencesManager.renameList(nameListsAdapter.getItem(position), newListName);
+                        nameListsAdapter.getContent().set(position, newListName);
+                        nameListsAdapter.notifyItemChanged(position);
+                    }
+                })
+                .show();
+    }
+
+    private void showDeleteDialog(final int position) {
```
Gear61: Same feedback as above. Let's turn this deletion dialog into another module that the activity contains and coordinates.

Gear61 (changes requested):
I know I'm requesting changes, but this is already fantastic. The test plan is great and the code is already good and very clean. But I want us to keep the highest possible standard for this project, so let's keep improving it!

mohammedhemaid:
There's a memory leak in the dialogs, I can't fix them :(

app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public interface Listener {
+        void onDeleteListConfirmed(int position);
+    }
+
+    private FragmentManager fragmentManager;
+    private Listener listener;
+    private int position;
+
+    public DeleteListDialog(FragmentManager fragmentManager, Listener listener) {
```
Gear61: We don't need a FragmentManager right? We should just contain an instance of MaterialDialog. It abstracts all that away for us.

app/build.gradle
```diff
@@ -56,4 +56,6 @@ dependencies {
     implementation 'com.nbsp:library:1.02'
     implementation 'com.github.deano2390:MaterialShowcaseView:1.2.0'
     implementation 'com.squareup:seismic:1.0.2'
+    //debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.0'
```
Gear61: nit: Comment this back in and delete the newline. I think when you compile release builds, it shouldn't include this, so we should be good.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -104,6 +106,13 @@ protected void onCreate(Bundle savedInstanceState) {
         setNoContent();
     }
 
+    @Override
+    protected void onStart() {
+        super.onStart();
+        renameListDialog = new RenameListDialog(getSupportFragmentManager(), this, preferencesManager);
```
Gear61: We can just do this in onCreate(), alongside everything else. This is better for consistency. Any particular reason you chose onStart()?

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -325,40 +334,35 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-                })
-                .show();
+        renameListDialog.show(position);
+    }
+
+    @Override
+    public void onRenameListConfirmed(int position, String newListName) {
+        dataSource.renameList(nameListsAdapter.getItem(position), newListName);
+        preferencesManager.renameList(nameListsAdapter.getItem(position), newListName);
+        nameListsAdapter.renameItem(position, newListName);
```
Gear61: renameItem() should do the notifyItemChanged() within it. Let's make MainActivity as much of a "coordinator" as possible, so it can just call powerful APIs and leave the more specific commands (like changing the list UI with notifyItemX) to individual modules.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -325,40 +334,35 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-                .show();
+        deleteListDialog.show(position);
+    }
+
+    @Override
+    public void onDeleteListConfirmed(int position) {
+        dataSource.deleteList(nameListsAdapter.getItem(position));
+        preferencesManager.removeNameList(nameListsAdapter.getItem(position));
+        nameListsAdapter.deleteItem(position);
+        nameListsAdapter.notifyItemRemoved(position);
```
Gear61: nit: Same feedback as above. Let's absorb notifyItemRemoved into the deleteItem() implementation.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -325,40 +334,35 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        setNoContent();
+    }
+
+    @Override
+    protected void onStop() {
+        super.onStop();
+        renameListDialog = null;
```
Gear61: Was this leaking in LeakCanary otherwise? I don't think we need these? MainActivity is also the base activity; if the user leaves it; the entire app is now dead.

Gear61 (changes requested):
Thanks for making these changes! One more round of changes, and we should be good.

app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
mohammedhemaid: Because I am extending the DialogFragment I need to pass the FragmentManager in the show method to use the context in onCreateDialog, if I make an instance of the material dialog inside the class it will make a NullPointerException.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
mohammedhemaid: not it's just leaking in the LeakCanary, moving from MainActivity to ListActivity makes a leak in leakCanary.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
mohammedhemaid: creating the dialogs in the onStart and clean them on OnStop, thinking this will reduce the memory leaks.

app/src/main/java/com/randomappsinc/studentpicker/home/RenameListDialog.java
```diff
@@ -0,0 +1,57 @@
...
+        this.listener = listener;
+        this.preferencesManager = preferencesManager;
+    }
+
+    @NonNull
+    @Override
+    public Dialog onCreateDialog(@Nullable Bundle savedInstanceState) {
+        return new MaterialDialog.Builder(getActivity())
+                .title(R.string.rename_list)
+                .input(getString(R.string.new_list_name), "", (dialog, input) -> {
```
mohammedhemaid: Can't pass an instance of Material Dialog to this class Because of the getString() need a context, we can pass it through the constructor of the RenameListDialog or we can keep using the Fragment manager

Gear61: We can pass a context in the constructor and this class can build the MaterialDialog itself. That abstracts away pretty much the entire implementation from the activity. You can find a similar-ish example to our use case here: https://github.com/Gear61/Random-Name-Picker/blob/master/app/src/main/java/com/randomappsinc/studentpicker/choosing/ChoicesDisplayDialog.java

app/src/main/java/com/randomappsinc/studentpicker/home/NameListsAdapter.java
```diff
@@ -77,42 +71,16 @@ String getItem(int position) {
...
-                    content.remove(position);
-                    notifyItemRemoved(position);
-                    delegate.setNoContent();
-                })
-                .show();
+    void deleteItem(int position) {
+        content.remove(position);
+        notifyItemRemoved(position);
+        delegate.setNoContent();
+
```
Gear61: nit: Delete the extra newline here.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -309,4 +329,18 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
         return super.onOptionsItemSelected(item);
     }
+
+    @Override
+    public void onRenameListConfirmed(int position, String newListName) {
```
Gear61: nit: Move this under onItemEditClick to keep similar code next to each other and improve readability.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -309,4 +329,18 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        dataSource.renameList(nameListsAdapter.getItem(position), newListName);
+        preferencesManager.renameList(nameListsAdapter.getItem(position), newListName);
+        nameListsAdapter.renameItem(position, newListName);
+    }
+
+    @Override
+    public void onDeleteListConfirmed(int position) {
```
Gear61: nit: Move this below onItemDeleteClick.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -64,7 +66,11 @@
     @BindString(R.string.list_duplicate) String listDuplicate;
 
     private PreferencesManager preferencesManager;
+    private DataSource dataSource;
     private NameListsAdapter nameListsAdapter;
+    private RenameListDialog renameListDialog;
+    private DeleteListDialog deleteListDialog;
+    private MaterialDialog.Builder materialDialogBuilder;
```
Gear61: We can just pass in context and the class can create MaterialDialogs itself.

Gear61 (changes requested):
Thanks for iterating on this! If you're still unclear on what I'm looking for, feel free to ping me on WhatsApp!

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
Gear61: Can you explain when this code leaks? Maybe we need to null out the dialog references in the dialog classes in onDestroy() or something...?

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
Gear61: Ah okay, that makes sense since ListActivity is not the base activity.

Gear61 (approved):
This is fantastic; thanks for doing this!

PR #78, opened 2019-12-20
Updating content of delete dialog to contain the name of the list

alaaalzibda:
Adding list name context to delete dialog

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
@@ -13,19 +13,21 @@ public class DeleteListDialog {
         void onDeleteListConfirmed(int position);
     }
 
-    private MaterialDialog dialog;
+    public MaterialDialog dialog;
     private int position;
 
     DeleteListDialog(@NonNull Listener listener, Context context) {
         dialog = new MaterialDialog.Builder(context)
                 .title(R.string.confirm_deletion_title)
-                .content(R.string.confirm_deletion_message)
                 .positiveText(android.R.string.yes)
                 .negativeText(android.R.string.cancel)
                 .onPositive((dialog, which) ->
                         listener.onDeleteListConfirmed(position)
                 ).build();
     }
+
+    // Instead of sending the content inside the show function, Defining the dialog object as public is better to get an access for it from the MainActivity
+    // I think changing the content inside a function called "show" isn't a good idea.
 
     public void show(int position) {
         this.position = position;
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
@@ -26,14 +26,11 @@ public class DeleteListDialog {
                 ).build();
     }
 
-    // Well there is no purpose of sending the position here do you keep it for the upcoming work?
     void presentForList(int position, String listName) {
         this.position = position;
-        String dialogContent = dialog.getContext().getString(R.string.confirm_deletion_message," \"" + listName + "\"");
+        String dialogContent =
+                dialog.getContext().getString(R.string.confirm_deletion_message, " \"" + listName + "\"");
         dialog.setContent(dialogContent);
-    }
-
-    public void show() {
         dialog.show();
     }
 }
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
@@ -13,7 +13,7 @@ public class DeleteListDialog {
         void onDeleteListConfirmed(int position);
     }
 
-    public MaterialDialog dialog;
+    private MaterialDialog dialog;
     private int position;
 
     DeleteListDialog(@NonNull Listener listener, Context context) {
@@ -26,8 +26,9 @@ public class DeleteListDialog {
                 ).build();
     }
 
-    // Instead of sending the content inside the show function, Defining the dialog object as public is better to get an access for it from the MainActivity
-    // I think changing the content inside a function called "show" isn't a good idea.
+    void setContent(String content) {
+        dialog.setContent(content);
+    }
 
     public void show(int position) {
         this.position = position;
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
@@ -28,12 +28,12 @@ public class DeleteListDialog {
 
     // Well there is no purpose of sending the position here do you keep it for the upcoming work?
     void presentForList(int position, String listName) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
```diff
@@ -13,20 +13,22 @@
...
                 .title(R.string.confirm_deletion_title)
-                .content(R.string.confirm_deletion_message)
                 .positiveText(android.R.string.yes)
                 .negativeText(android.R.string.cancel)
                 .onPositive((dialog, which) ->
                         listener.onDeleteListConfirmed(position)
                 ).build();
     }
 
+    // Instead of sending the content inside the show function, Defining the dialog object as public is better to get an access for it from the MainActivity
```
Gear61: Hmm, this is a good point that setting the content within this method violates single responsibility design. However, exposing the dialog breaks encapsulation since anyone can now mutate the dialog and this mutation is now increasing the complexity of the activity, which we want to avoid. What if we made the method presentForList(int position, String listName)?

app/src/main/res/values/edit_name_list_strings.xml
```diff
@@ -27,7 +27,7 @@
     <string name="import_names_from_list">Import names from another list(s)</string>
 
     <string name="confirm_deletion_title">Confirm List Deletion</string>
-    <string name="confirm_deletion_message">Are you sure you want to delete this name list?</string>
+    <string name="confirm_deletion_message">Are you sure you want to delete your name list</string>
```
Gear61: This makes the code a little longer, but I think this is better separation of responsibilities. We want to put as much of the text into the XML as possible.

```suggestion
    <string name="confirm_deletion_message">Are you sure you want to delete your name list %s?</string>
```

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -189,7 +189,9 @@ public void onRenameListConfirmed(int position, String newListName) {
     }
 
     @Override
-    public void onItemDeleteClick(int position) {
+    public void onItemDeleteClick(int position, String listName) {
+        String dialogContent = getString(R.string.confirm_deletion_message) + " \"" + listName + "\"?";
```
Gear61: Once we start using placeholders, the getString() API can actually support it via additional parameters; it's pretty cool.

```suggestion
        String dialogContent = getString(R.string.confirm_deletion_message, " \"" + listName + "\");
```

Gear61 (changes requested):
This is a good start and I really appreciate the comment about the show() API! Also, can you attach a screenshot that shows the new changes working? All pull requests should have a test plan that proves they work to ensure the quality of the code.

app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
alaaalzibda: Screenshot_20191221-105447
Screenshot_20191221-105450

alaaalzibda: Thanks for the feedback, but I can't see a function called presentForList(int position, String listName)?
I created a setter function inside the DeleteListDialog and defined MaterialDialog as private again.

app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
```diff
@@ -19,14 +19,17 @@
...
                 .onPositive((dialog, which) ->
                         listener.onDeleteListConfirmed(position)
                 ).build();
     }
 
+    void setContent(String content) {
+        dialog.setContent(content);
+    }
+
     public void show(int position) {
```
Gear61: Sorry for the miscommunication, I meant change this API to presentForList(int position, String listName). You can also do dialog.getContext() to get a context, and you can call getString from there. I think it's better to couple these 2 APIs together, since there isn't a use-case where you would call 1 without the other.

Gear61 (changes requested):
Thanks for iterating on the feedback! One more change and we should be good to go. My main goal here is to keep MainActivity as slim as possible.

app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
```diff
@@ -19,16 +19,21 @@
...
-                .content(R.string.confirm_deletion_message)
                 .positiveText(android.R.string.yes)
                 .negativeText(android.R.string.cancel)
                 .onPositive((dialog, which) ->
                         listener.onDeleteListConfirmed(position)
                 ).build();
     }
 
-    public void show(int position) {
+    // Well there is no purpose of sending the position here do you keep it for the upcoming work?
```
Gear61: I'm glad you're asking these questions! For the future, questions like these are better left as comments on your own pull request instead of in the code (since you will have to delete them later on).

The position is unfortunately a bit of a hack so the activity knows which item to update in the adapter once the action is taken.

app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
```diff
@@ -19,16 +19,21 @@
...
+    void presentForList(int position, String listName) {
         this.position = position;
+        String dialogContent = dialog.getContext().getString(R.string.confirm_deletion_message," \"" + listName + "\"");
+        dialog.setContent(dialogContent);
+    }
+
+    public void show() {
```
Gear61: nit: Merge the show() into presentForList().

app/src/main/java/com/randomappsinc/studentpicker/home/DeleteListDialog.java
```diff
@@ -19,16 +19,21 @@
...
                 .onPositive((dialog, which) ->
                         listener.onDeleteListConfirmed(position)
                 ).build();
     }
 
-    public void show(int position) {
+    // Well there is no purpose of sending the position here do you keep it for the upcoming work?
+    void presentForList(int position, String listName) {
         this.position = position;
+        String dialogContent = dialog.getContext().getString(R.string.confirm_deletion_message," \"" + listName + "\"");
```
Gear61: nit: Slight formatting improvement

```suggestion
        String dialogContent = dialog.getContext().getString(R.string.confirm_deletion_message, " \"" + listName + "\"");
```

Gear61 (changes requested):
So close! Feel free to ping me on Slack if you aren't sure what I mean when I'm requesting changes. It's better to solve the issues proactively instead of committing the code and then having to change it.

ALAA999:
Did you see the commit description?
"Alaa Alzibda authored and Alaa Alzibda committed 6 minutes ago" hhhh
Seriously, what the hell?! :)

ALAA999:
Well, I realized that I changed the author name, that's why git shows the another and the committer.
But how do I return it to normal? showing only the committer name?

Gear61 (approved):
This is great! Thanks for improving the app!

PR #83, opened 2019-12-26
Add the empty state message and button

mohammedhemaid:
The video of the Feature:
https://photos.app.goo.gl/at8yKoig94pti1749

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
@@ -4,10 +4,15 @@ import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.TextView;
 
 import androidx.fragment.app.Fragment;
 
 import com.randomappsinc.studentpicker.R;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+import butterknife.Unbinder;
 
 public class GroupsFragment extends Fragment {
 
@@ -15,9 +20,23 @@ public class GroupsFragment extends Fragment {
         return new GroupsFragment();
     }
 
+    @BindView(R.id.no_groups) TextView noGroups;
+
+    private Unbinder unbinder;
+
     @Override
-    public View onCreateView(LayoutInflater inflater, ViewGroup container,
-                             Bundle savedInstanceState) {
-        return inflater.inflate(R.layout.fragment_groups, container, false);
+    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
+        View rootView = inflater.inflate(R.layout.fragment_groups, container, false);
+        unbinder = ButterKnife.bind(this,rootView);
+
+        noGroups.setVisibility(View.VISIBLE);
+
+        return rootView;
+    }
+
+    @Override
+    public void onDestroyView() {
+        super.onDestroyView();
+        unbinder.unbind();
     }
 }
--- a/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
@@ -1,6 +1,5 @@
 package com.randomappsinc.studentpicker.grouping;
 
-
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
--- a/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
@@ -1,0 +1,24 @@
+package com.randomappsinc.studentpicker.grouping;
+
+
+import android.os.Bundle;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+import androidx.fragment.app.Fragment;
+
+import com.randomappsinc.studentpicker.R;
+
+public class GroupsFragment extends Fragment {
+
+    public static GroupsFragment getInstance() {
+        return new GroupsFragment();
+    }
+
+    @Override
+    public View onCreateView(LayoutInflater inflater, ViewGroup container,
+                             Bundle savedInstanceState) {
+        return inflater.inflate(R.layout.fragment_groups, container, false);
+    }
+}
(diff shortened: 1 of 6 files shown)


mohammedhemaid:
it's very weird, why my branch is telling me that I am behind, it should work well.
I've tried this:
https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch
But it didn't work.
Maybe it will solve it by itself like the first time :)
Also, I am thinking to create a new pull request

alaaalzibda:
So basically the problem is because you forked the repo a while ago, then Alex made some changes on the code and pushed these changes, now they are pushed to his repo (the main one) not yours.
Your repo still has the old version you forked!
You have to call this:
 - git remote add upstream https://github.com/Gear61/Random-Name-Picker.git
 - git fetch upstream
 - git rebase upstream/master
 - git push origin master --force

So this will get the latest version of Alex's coed and pushes it to your repo.
https://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository
I hope this will help :)

mohammedhemaid:
@ALAA999 
No, it's not because of that, this branch was updated by me and there are no commits form Alex

alaaalzibda:
So you have the latest version of Alex's repo and it tells you that you are x commits behind?

mohammedhemaid:
@ALAA999  
Yes, it's like that.

alaaalzibda:
That's really weird, I had this problem before, I did what I told you before then it worked!
well if nothing works with you, just make a copy of the files you changed (save them somewhere else) then delete the repo you forked and fork it again then add your changes locally and push it.
But make it your last option, I think we should wait for Alex, he might help us.

Gear61:
Hey guys, sorry I'm late here. I've been busy during the holidays. My theory is that this pull request (PR) was submitted before I merged the first pull request in to update group-maker in this repo, but I don't think that's the case since it would be weird to have another PR out while the previous one was still open. I feel like this is just a shortcoming of GitHub. Anyways, the PR looks good to me now, so I'm going to review it!

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
```diff
@@ -4,20 +4,39 @@
...
+    @BindView(R.id.no_groups) TextView noGroups;
+
+    private Unbinder unbinder;
+
+    @Override
+    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
+        View rootView = inflater.inflate(R.layout.fragment_groups, container, false);
+        unbinder = ButterKnife.bind(this,rootView);
+
+        noGroups.setVisibility(View.VISIBLE);
```
Gear61, line 32: We should do this in XML. We should separate layout logic into the XML and out of the Java as much as possible to focus their responsibilities.

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListTabsAdapter.java
```diff
@@ -9,37 +9,35 @@
...
-        return fragmentList.get(position);
+        switch (position) {
+            case 0:
+                return NameChoosingFragment.getInstance(listName);
+            case 1:
+                return GroupsFragment.getInstance();
+            case 2:
+                return EditNameListFragment.getInstance(listName);
+            default:
+                throw new IllegalArgumentException("There should only be 3 tabs!");
```
Gear61, line 39: Just curious as to why you reverted back to this approach. I'm personally fine either way. I think I did this way originally so I never had to store references to fragments.

Gear61 (changes requested):
Thanks for the great test plan!

I also just realized that we should set the offscreenPageLimit to 2 on the ViewPager, since we have 3 tabs now. This way, all the fragments will be alive all the time no matter which tab you're on. With the current implementation, I think if you go from "Choose" to "Edit", the "Choose" fragment will die and will need to be recreated when the user tabs back. Our app is very light, so I think this is okay.

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListTabsAdapter.java
mohammedhemaid: After I read this, https://proandroiddev.com/the-seven-actually-10-cardinal-sins-of-android-development-491d2f64c8e0 
it says it's not good to do it like this.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
mohammedhemaid: it's just a temporary line, I will remove it after complete the scenario anyways, Should I put it in the XML now.

app/src/main/java/com/randomappsinc/studentpicker/listpage/ListTabsAdapter.java
Gear61: Ah okay, that makes sense. Looks like my initial inclination was right. I have learned several times over the years that you don't mess with fragments. :D

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsFragment.java
Gear61: Ah, you did it for the demo. Sounds good, we can keep it for now. :)

app/src/main/res/values/strings.xml
```diff
@@ -16,6 +16,10 @@
         <item>@string/edit</item>
     </array>
     <string name="options_for">Options for:&#160;</string>
+    <string name="make_groups">Make Groups</string>
+    <string name="no_groups_created_message">
+        Use the {ion-android-settings} icon to customize your groups
+        and then tap \"Make Groups\" at the bottom</string>
```
Gear61: nit: Small formatting change so developers can focus on the string and adding a period since this is a complete sentence.

```suggestion
        and then tap \"Make Groups\" at the bottom.
    </string>
```

Gear61 (changes requested):
Can you make the ViewPager change as well? I feel like it counts towards new fragment setup. Also, really appreciate you digging into the fragment references issue! Taking the extra time to do the right thing is a crucial behavior for strong senior engineers.

mohammedhemaid:
The video for the app:
https://photos.app.goo.gl/eZzFbde9zEAqJvqVA

Gear61 (approved):
Awesome job @mohammedhemaid!

PR #86, opened 2019-12-29
Prefill change name dialog with the name selected

alaaalzibda:


Gear61 (approved):
Glad this was a 1 liner. Thanks for the change!

PR #88, opened 2019-12-31
Voice input for edit name screen, setting sliding animation for speec…

alaaalzibda:
…h dialog
One thing I want to mention is that exitAnim when sliding out the dialog not working, however I set both for attributes overridePendingTransition (enterAnim, exitAnim)
I searched for it most results were telling me that I have to overridePendingTransition after the finish call of the intent, but we don't call finish for our dialog.

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
@@ -76,7 +76,7 @@ public class EditNameListFragment extends Fragment implements
 
         newNameInput.setHint(R.string.name_hint);
         newNameInput.setInputType(InputType.TYPE_TEXT_FLAG_CAP_WORDS);
-        newNameInput.setAdapter(new NameCreationACAdapter(getActivity()));
+        newNameInput.setAdapter(new NameCreationAutoCompleteAdapter(getActivity()));
         plus.setImageDrawable(new IconDrawable(
                 getActivity(),
                 IoniconsIcons.ion_android_add).colorRes(R.color.white));
--- a/app/src/main/java/com/randomappsinc/studentpicker/editing/NameCreationAutoCompleteAdapter.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/editing/NameCreationAutoCompleteAdapter.java
@@ -15,27 +15,24 @@ import com.randomappsinc.studentpicker.database.DataSource;
 import java.util.ArrayList;
 import java.util.List;
 
-import butterknife.BindView;
-import butterknife.ButterKnife;
-
-public class NameCreationACAdapter extends BaseAdapter implements Filterable {
+public class NameCreationAutoCompleteAdapter extends BaseAdapter implements Filterable {
 
     private List<String> suggestions;
     private DataSource dataSource;
 
     private Context context;
 
-    public NameCreationACAdapter(Context context) {
+    NameCreationAutoCompleteAdapter(Context context) {
         this.context = context;
         this.dataSource = new DataSource(context);
         this.suggestions = new ArrayList<>();
     }
 
     static class ViewHolder {
-        @BindView(R.id.suggestion) TextView suggestion;
+        TextView suggestionTextView;
 
         ViewHolder(View view) {
-            ButterKnife.bind(this, view);
+            suggestionTextView = (TextView) view;
         }
     }
 
@@ -64,7 +61,7 @@ public class NameCreationACAdapter extends BaseAdapter implements Filterable {
         } else {
             holder = (ViewHolder) view.getTag();
         }
-        holder.suggestion.setText(getItem(position));
+        holder.suggestionTextView.setText(getItem(position));
         return view;
     }
 
--- a/app/src/main/res/layout/name_suggestion_cell.xml
+++ b/app/src/main/res/layout/name_suggestion_cell.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <TextView
     xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/suggestion"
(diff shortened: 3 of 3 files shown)


alaaalzibda:
Screenshot_20191231-155503
Screenshot_20191231-155533

app/src/main/res/layout/lists_with_add_content.xml
```diff
@@ -59,6 +59,20 @@
                 android:layout_width="match_parent"
                 android:layout_height="44dp"
                 />
+
+            <com.joanzapata.iconify.widget.IconTextView
+                android:id="@+id/voice_entry_icon"
```
Gear61: Let's move all these XML attributes into a reusable style.

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -120,6 +121,11 @@ public void showNameOptions(final int position) {
         nameEditChoicesDialog.showChoices(namesAdapter.getItem(position));
     }
 
+    @OnClick(R.id.voice_entry_icon)
+    public void voiceEntry() {
+        SpeechUtil.openSpeechToTextDialog(getActivity());
```
Gear61: Hmmm, I just realized that using the activity isn't great since we should use the startActivity of the fragment. This will allow us to keep the onActivityResult block in the fragment where the logic rightfully belongs. Let's change the utility method to only construct the intent.

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
```diff
@@ -0,0 +1,45 @@
...
+        intent.putExtra(RecognizerIntent.EXTRA_PROMPT, activity.getString(R.string.list_name_input_speech_message));
+        try {
+            activity.startActivityForResult(intent, SPEECH_REQUEST_CODE);
+            activity.overridePendingTransition(R.anim.slide_in_up, R.anim.slide_out_up);
+        } catch (ActivityNotFoundException exception) {
+            UIUtils.showLongToast(R.string.speech_not_supported, activity);
+        }
+    }
+
+    public static String processSpeechResults(int resultCode, Intent data, Context context) {
```
Gear61: We should mark this method as @Nullable so developers are aware that this can return null.
```suggestion
    @Nullable
    public static String processSpeechResults(int resultCode, Intent data, Context context) {
```

app/src/main/res/anim/slide_in_up.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Gear61: Let's do this instead:
1. Don't use an Android resource, which we don't have control over
2. Separate the /> into its own line to prevent merge conflicts
3. Move namespace to a separate line for cleanliness
4. Rename this to slide_in_from_bottom, since that's more descriptive
```
<?xml version="1.0" encoding="utf-8"?>
<translate
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:fromYDelta="100%p"
    android:toYDelta="0%p"
    android:duration="@integer/default_anim_length"
    />
```

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
```diff
@@ -0,0 +1,45 @@
...
+    public static final int SPEECH_REQUEST_CODE = 1;
+
+    public static void openSpeechToTextDialog(Activity activity) {
+        Intent intent = new Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH);
+        intent.putExtra(RecognizerIntent.EXTRA_LANGUAGE_MODEL, RecognizerIntent.LANGUAGE_MODEL_FREE_FORM);
+        intent.putExtra(RecognizerIntent.EXTRA_LANGUAGE, Locale.getDefault());
+        intent.putExtra(RecognizerIntent.EXTRA_PROMPT, activity.getString(R.string.list_name_input_speech_message));
+        try {
+            activity.startActivityForResult(intent, SPEECH_REQUEST_CODE);
+            activity.overridePendingTransition(R.anim.slide_in_up, R.anim.slide_out_up);
```
Gear61: I think you misunderstood how this API works similar to me. The first animation refers to how the new activity will animate in. The second refers to how the current activity will animate out. Since the speech dialog goes over the invoking activity, the current activity shouldn't move. This isn't the cleanest solution, but the way I made the current activity stay was to create a "stay.xml" animation like this:

```
<?xml version="1.0" encoding="utf-8"?>
<translate
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:duration="@integer/default_anim_length"
    android:fromYDelta="0%p"
    android:toYDelta="0%p"
    />
```

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
```diff
@@ -0,0 +1,45 @@
...
+import com.randomappsinc.studentpicker.R;
+
+import java.util.List;
+import java.util.Locale;
+
+import static android.app.Activity.RESULT_OK;
+
+public class SpeechUtil {
+
+    public static final int SPEECH_REQUEST_CODE = 1;
```
Gear61: We should let the callers customize this code IMHO. If the caller is using 1 for something else, we will have bugs :(

Gear61 (changes requested):
This is a good start! Also, there are merge conflicts between your master and mine, probably because I fixed several bugs last night and updated master. You should update your master against mine; let me know if you need help with that!

app/src/main/res/layout/lists_with_add_content.xml
alaaalzibda: That sounds good but why not we make a separate layout and include wherever we want using the include tag?

Issue #85, opened 2019-12-29
[Feature] Voice input for names


Gear61:
Typing on a mobile phone sucks, so I try to have this feature in as many places as I can. Similar to the homepage, let's add a microphone icon to the name adding page, so the user can "type" with their voice in a single tap. For this task, let's keep reusability in mind.

Notes:
- Speech prompt message should be "Tell us what name you want to add"
- Let's move the code that starts the Google speech dialog into a new class called SpeechUtil. The method should be called "openSpeechToTextDialog"
- Let's move the code that processes the returned speech result intent into the new SpeechUtil class as well. We can call it "processSpeechResults". It will return null if no speech was detected and a string otherwise
- Let's move the microphone button XML styling attributes into a reusable style
- BONUS: In the utility method to open the Google speech dialog, make the entry transition for it be a slide up from the bottom. It exits out by sliding down through the bottom, so we should make these symmetrical.

Gear61:
I think you're ready to take on a harder issue now @ALAA999, so here it is. Let me know if you need any help with this!

Gear61:
This is on prod!

PR #89, opened 2020-01-01
Voice input for edit name screen, setting sliding animation for speec…

alaaalzibda:
…h dialog

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
@@ -1,5 +1,7 @@
 package com.randomappsinc.studentpicker.editing;
 
+import android.content.ActivityNotFoundException;
+import android.content.Intent;
 import android.os.Bundle;
 import android.text.InputType;
 import android.view.LayoutInflater;
@@ -23,6 +25,7 @@ import com.randomappsinc.studentpicker.database.DataSource;
 import com.randomappsinc.studentpicker.database.NameListDataManager;
 import com.randomappsinc.studentpicker.home.MainActivity;
 import com.randomappsinc.studentpicker.models.ListInfo;
+import com.randomappsinc.studentpicker.utils.SpeechUtil;
 import com.randomappsinc.studentpicker.utils.UIUtils;
 
 import java.util.List;
@@ -36,6 +39,8 @@ import butterknife.Unbinder;
 public class EditNameListFragment extends Fragment implements
         NameEditChoicesDialog.Listener, RenameDialog.Listener, DeleteNameDialog.Listener,
         DuplicationDialog.Listener, MergeNameListsDialog.Listener {
+
+    private static final int SPEECH_REQUEST_CODE = 4;
 
     public static EditNameListFragment getInstance(String listName) {
         Bundle bundle = new Bundle();
@@ -118,6 +123,17 @@ public class EditNameListFragment extends Fragment implements
     @OnItemClick(R.id.content_list)
     public void showNameOptions(final int position) {
         nameEditChoicesDialog.showChoices(namesAdapter.getItem(position));
+    }
+
+    @OnClick(R.id.voice_entry_icon)
+    public void voiceEntry() {
+        Intent intent = SpeechUtil.openSpeechToTextDialog(getActivity());
+        try {
+            startActivityForResult(intent, SPEECH_REQUEST_CODE);
+            getActivity().overridePendingTransition(R.anim.slide_in_from_bottom, R.anim.stay);
+        } catch (ActivityNotFoundException exception) {
+            UIUtils.showLongToast(R.string.speech_not_supported, getActivity());
+        }
     }
 
     @Override
@@ -207,5 +223,14 @@ public class EditNameListFragment extends Fragment implements
                 return true;
         }
         return super.onOptionsItemSelected(item);
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        if (requestCode == SPEECH_REQUEST_CODE) {
+            String searchInput = SpeechUtil.processSpeechResults(resultCode, data, getActivity());
+            newNameInput.setText(searchInput);
+        }
     }
 }
--- a/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
@@ -230,8 +230,9 @@ public class EditNameListFragment extends Fragment implements
         super.onActivityResult(requestCode, resultCode, data);
         if (requestCode == SPEECH_REQUEST_CODE) {
             String searchInput = SpeechUtil.processSpeechResult(resultCode, data, getActivity());
-            if (searchInput != null)
+            if (searchInput != null){
                 newNameInput.setText(searchInput);
+            }
         }
     }
 }
--- a/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
@@ -40,7 +40,7 @@ public class EditNameListFragment extends Fragment implements
         NameEditChoicesDialog.Listener, RenameDialog.Listener, DeleteNameDialog.Listener,
         DuplicationDialog.Listener, MergeNameListsDialog.Listener {
 
-    private static final int SPEECH_REQUEST_CODE = 4;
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -37,6 +40,8 @@
         NameEditChoicesDialog.Listener, RenameDialog.Listener, DeleteNameDialog.Listener,
         DuplicationDialog.Listener, MergeNameListsDialog.Listener {
 
+    private static final int SPEECH_REQUEST_CODE = 4;
```
Gear61: nit: We can just use 1 here.

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
```diff
@@ -0,0 +1,38 @@
...
+import androidx.annotation.Nullable;
+
+import com.randomappsinc.studentpicker.R;
+
+import java.util.List;
+import java.util.Locale;
+
+import static android.app.Activity.RESULT_OK;
+
+public class SpeechUtil {
```
Gear61, line 16: nit: Add a newline for separation.

```suggestion
public class SpeechUtil {

```

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
```diff
@@ -0,0 +1,38 @@
...
+    public static Intent openSpeechToTextDialog(Context context) {
+        Intent intent = new Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH);
+        intent.putExtra(RecognizerIntent.EXTRA_LANGUAGE_MODEL, RecognizerIntent.LANGUAGE_MODEL_FREE_FORM);
+        intent.putExtra(RecognizerIntent.EXTRA_LANGUAGE, Locale.getDefault());
+        intent.putExtra(RecognizerIntent.EXTRA_PROMPT, context.getString(R.string.list_name_input_speech_message));
```
Gear61: We need to make this customizable. So instead of taking in a context, take in a string directly and use it here.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -271,33 +274,12 @@ public void onRequestPermissionsResult(
...
-                }
-
-                List<String> result = data.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS);
-                if (result == null || result.isEmpty()) {
-                    UIUtils.showLongToast(R.string.speech_unrecognized, this);
-                    return;
-                }
-                String searchInput = result.get(0);
+                String searchInput = SpeechUtil.processSpeechResults(resultCode, data, this);
                 newListInput.setText(searchInput);
```
Gear61: We should only call setText() if the return value from processSpeechResults is non-null. Otherwise if the user opts-out of the speech dialog (just tap outside of it), the text input will be incorrectly cleared I think.

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.randomappsinc.studentpicker.R;
+
+import java.util.List;
+import java.util.Locale;
+
+import static android.app.Activity.RESULT_OK;
+
+public class SpeechUtil {
+    public static Intent openSpeechToTextDialog(Context context) {
```
Gear61: We should update the name to match the new logic.
```suggestion
    public static Intent getSpeechToTextIntent(Context context) {
```

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
```diff
@@ -0,0 +1,38 @@
...
+        Intent intent = new Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH);
+        intent.putExtra(RecognizerIntent.EXTRA_LANGUAGE_MODEL, RecognizerIntent.LANGUAGE_MODEL_FREE_FORM);
+        intent.putExtra(RecognizerIntent.EXTRA_LANGUAGE, Locale.getDefault());
+        intent.putExtra(RecognizerIntent.EXTRA_PROMPT, context.getString(R.string.list_name_input_speech_message));
+        return intent;
+    }
+
+    @Nullable
+    public static String processSpeechResults(int resultCode, Intent data, Context context) {
```
Gear61: nit: I feel like this is slightly better.
```suggestion
    public static String processSpeechResult(int resultCode, Intent data, Context context) {
```

Gear61 (changes requested):
Looks way better than before! Keep it up!

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
alaaalzibda: You avoid defining it as public static in SpeechUtil file to prevent some kind of hackers access it or from other files?

app/src/main/java/com/randomappsinc/studentpicker/utils/SpeechUtil.java
alaaalzibda: I think sending the string itself it better because this function is called from MainActivity and EditNameFragment, both have different implementation.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -271,34 +274,14 @@ public void onRequestPermissionsResult(
...
-
-                List<String> result = data.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS);
-                if (result == null || result.isEmpty()) {
-                    UIUtils.showLongToast(R.string.speech_unrecognized, this);
-                    return;
-                }
-                String searchInput = result.get(0);
-                newListInput.setText(searchInput);
+                String searchInput = SpeechUtil.processSpeechResult(resultCode, data, this);
+                if (searchInput != null)
```
Gear61: nit: Please add the brackets for consistency with the rest of the codebase.
```suggestion
                if (searchInput != null) {
```

app/src/main/res/values/strings.xml
```diff
@@ -64,6 +64,7 @@
     <string name="delete_list">Delete List</string>
 
     <string name="list_name_input_speech_message">Tell us what you want to name your list</string>
+    <string name="name_input_speech_message_to_list">Tell us the name you want to add to %s\'s list</string>
```
Gear61: Can you change this to "Tell us what name you want to add" as mentioned in the issue: https://github.com/Gear61/Random-Name-Picker/issues/85

I think the list you're within is pretty clear since it's in the toolbar. It's not like when you clicked on a specific item within the list, of which there are many in the UI.

Also, there should be a edit_name_list_strings.xml file which is a better place for this resource.

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -208,4 +224,14 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
         return super.onOptionsItemSelected(item);
     }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        if (requestCode == SPEECH_REQUEST_CODE) {
+            String searchInput = SpeechUtil.processSpeechResult(resultCode, data, getActivity());
+            if (searchInput != null)
```
Gear61: nit: Please add the brackets for consistency.
```suggestion
            if (searchInput != null) {
```

Gear61 (changes requested):
Almost there! Thanks for iterating on the feedback!

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
alaaalzibda: Okey, I usually remove it to get less number of code lines.

app/src/main/res/values/strings.xml
alaaalzibda: Sorry, after moving to the next notes you asked for and working with the other stuff like animation, I forgot the first note which is about the speech promote message text :)

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
Gear61: Less code is good, but I feel like this trick makes the code look more like Python than Java. Also, the rest of the codebase is already using the brackets; it's best to be consistent.

Gear61 (approved):
This looks great! Thanks for making the changes!

PR #93, opened 2020-01-11
Add the settings dialog

mohammedhemaid:
The video of the Feature:
https://photos.app.goo.gl/kGMRyYctMPfQjt3W8


app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingFragment.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.joanzapata.iconify.fonts.IoniconsIcons;
+import com.randomappsinc.studentpicker.R;
+import com.randomappsinc.studentpicker.utils.UIUtils;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+import butterknife.Unbinder;
+
+public class GroupingFragment extends Fragment {
```
mohammedhemaid: I've changed the name of the GroupFragment to GroupingFragment

Gear61: nit: How about GroupMakingFragment? I feel like that's more declarative.

mohammedhemaid: yes, that is better.

app/src/main/res/values/grouping_strings.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Gear61, line 1: Thanks for proactively moving these into a separate file :)

app/src/main/res/menu/groups_menu.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
```
Gear61: nit: To prevent future merge conflicts
```suggestion
    xmlns:app="http://schemas.android.com/apk/res-auto"
    >
```

app/src/main/res/menu/groups_menu.xml
```diff
@@ -0,0 +1,10 @@
...
+
+    <item
+        android:id="@+id/settings"
+        android:title="@string/settings"
+        app:showAsAction="ifRoom"/>
```
Gear61: nit
```suggestion
        app:showAsAction="always"
        />
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,68 @@
...
+
+    void applySettings() {
+        applyNumberOfNames(namesInGroup);
+        applyNumberOfGroups(numGroups);
+    }
+
+    private void applyNumberOfNames(EditText namesInGroup) {
+        String numOfNamesInGroup = namesInGroup.getText().toString();
+        if (numOfNamesInGroup.isEmpty()) {
+            namesInGroup.setText("2");
```
Gear61: nit: We can use the string resource here. Hard-coding a "magic" string here puts us at risk of someone changing the value in the future and not this use.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,68 @@
...
+        numGroups.setText(String.valueOf(settings.getNumOfGroups()));
+        numGroups.clearFocus();
+    }
+
+    void applySettings() {
+        applyNumberOfNames(namesInGroup);
+        applyNumberOfGroups(numGroups);
+    }
+
+    private void applyNumberOfNames(EditText namesInGroup) {
```
Gear61: This method is private and takes in an EditText it has access to itself? Let's just remove the parameter and mutate the member variable directly instead? Seems simpler.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,68 @@
...
+        String numOfNamesInGroup = namesInGroup.getText().toString();
```
Gear61: nit: I know that the inputType is digits only, but I feel like being extra safe here doesn't hurt.
```suggestion
        String numOfNamesInGroup = namesInGroup.getText().toString().trim();
```

app/src/main/res/values/grouping_strings.xml
```diff
@@ -0,0 +1,13 @@
...
+<resources>
+    <string name="make_groups">Make Groups</string>
+    <string name="no_groups_created_message">
+        Use the {ion-android-settings} icon to customize your groups
+        and then tap \"Make Groups\" at the bottom.
+    </string>
+    <string name="grouping_setting">Grouping Setting</string>
+    <string name="default_number_of_names_in_each_group">2</string>
+    <string name="default_number_of_groups">1</string>
+    <string name="num_of_names_in_group"># of names in group:</string>
```
Gear61: nit: I feel like this is slightly better.
```suggestion
    <string name="num_of_names_in_group"># of names per group:</string>
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,68 @@
...
+        if (numOfNamesInGroup.isEmpty()) {
+            namesInGroup.setText("2");
+            settings.setNumOfNamesInGroup(2);
+        } else {
+            int userNumNames = Integer.parseInt(namesInGroup.getText().toString());
+            if (userNumNames <= 0) {
```
Gear61: It feels inconsistent to have the 2 error flows diverge (empty input vs. negative input) and do 2 different things. I think our logic should be like this:

1. Try to convert the string input into a number. If it's empty, make it 0. So it would be something like this:
`int inputtedNumber = numOfNamesInGroup.isEmpty() ? 0 : Integer.parseInt(namesInGroup.getText().toString());`
2. Check if the number from step #1 is <= 0, if so, run error recovery flow. And that error recovery flow would put us back at 2.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,68 @@
...
+import android.widget.EditText;
+
+import com.randomappsinc.studentpicker.R;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+
+class GroupingSettingsViewHolder {
+
+    @BindView(R.id.num_of_names_in_group) EditText namesInGroup;
```
Gear61: Sorry for the big change, but can we change the ID of the view to num_of_names_per_group and change all the variable naming to reflect that?

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsDialog.java
```diff
@@ -0,0 +1,41 @@
...
+                .onPositive((dialog, which) -> {
+                    settingsHolder.applySettings();
+                    UIUtils.showShortToast(R.string.settings_applied, context);
+                })
+                .onNegative((dialog, which) -> settingsHolder.revertSettings())
+                .cancelable(false)
+                .build();
+    }
+
+    void setSettingsHolder(GroupingSettingsViewHolder settingsHolder) {
```
Gear61: The architecture I have in NameChoosingFragment is bad, haha. Let's not take in a ViewHolder from outside. The dialog should just instantiate and hold the ViewHolder itself. The fragment should be unaware of this ViewHolder abstraction since it's a fragment and shouldn't be aware of details.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettings.java
```diff
@@ -0,0 +1,28 @@
+package com.randomappsinc.studentpicker.grouping;
+
+public class GroupingSettings {
+
+    private int numOfNamesInGroup;
+    private int numOfGroups;
+
+    public GroupingSettings() {
+        this.numOfNamesInGroup = 2;
```
Gear61: Having all these hardcoded values is kind of awkward. Let's move this 2 and 1 into integers.xml. Since GroupingSettings is just a model, it can take in 2 ints in the constructor for initial values. The fragment will fetch the default values (since it has a context) and pass them into the constructor.

app/src/main/res/values/grouping_strings.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="make_groups">Make Groups</string>
+    <string name="no_groups_created_message">
+        Use the {ion-android-settings} icon to customize your groups
+        and then tap \"Make Groups\" at the bottom.
+    </string>
+    <string name="grouping_setting">Grouping Setting</string>
```
Gear61:
```suggestion
    <string name="group_settings_dialog_title">Group Making Settings</string>
```

app/src/main/res/layout/group_making_settings.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical"
+    android:background="@color/white"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    >
+
```
Gear61, line 9: As I was thinking about this feature in my head, I realized that the number of names currently in the list would be helpful to the user to help figure out proper group settings. So let's put a TextView at the top that says, "This list has X names." where X is the number of names in the list. textSize of 17dp and textColor of dark_gray should be good...?

Gear61 (changes requested):
This is a great start and thanks for the demo! This is where things start getting hard, so your work is very much appreciated :)

app/src/main/res/values/integers.xml
```diff
@@ -2,4 +2,7 @@
 <resources>
     <integer name="default_anim_length">250</integer>
     <integer name="shorter_anim_length">150</integer>
+    <integer name="default_number_of_names_per_group">2</integer>
+    <integer name="default_number_of_groups">1</integer>
+
```
mohammedhemaid: sorry for this empty line I've deleted it and will be pushed in the next commit.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,69 @@
...
+import butterknife.BindView;
+import butterknife.ButterKnife;
+
+class GroupingSettingsViewHolder {
+
+    @BindView(R.id.number_of_names_in_list) TextView numberOfNamesInList;
+    @BindView(R.id.num_of_names_per_group) EditText namesPerGroup;
+    @BindView(R.id.num_of_groups) EditText numGroups;
+
+    private Context context;
```
Gear61: Let's not store a context here, every view exposes a getContext() API and this class has a bunch of views. It's dangerous holding contexts because they are often times activities underneath and if you don't manage the context variable properly, you can leak an entire activity and have a bad time.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,69 @@
...
+    private GroupingSettings settings;
+
+    GroupingSettingsViewHolder(View view, Context context, GroupingSettings settings) {
+        ButterKnife.bind(this, view);
+        this.context = context;
+        this.settings = settings;
+        numberOfNamesInList.setText(
```
Gear61: Is the plan to replace 0 with an actual number later? I'm fine doing this in a later pull request, just wanted to make sure.

Gear61 (changes requested):
Massive improvement over the last iteration, only have the context feedback to resolve. Thanks for doing this!

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
mohammedhemaid: yes, this is a reason for passing the context, you know in the dialog issue we spend a long time in it actually, I hate to pass the context because of the leak but I couldn't find a better way to set the text so, I kept it like this for now.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,69 @@
...
+        applyNumberOfGroups();
+    }
+
+    private void applyNumberOfNames() {
+        String numOfNamesPerGroup = namesPerGroup.getText().toString();
+        int inputtedNumber = numOfNamesPerGroup.isEmpty() ? 0 :
+                Integer.parseInt(namesPerGroup.getText().toString());
+        if (inputtedNumber <= 0) {
+            namesPerGroup.setText(R.string.default_number_of_names_per_group);
+            settings.setNumOfNamesPerGroup(context.getResources().getInteger(R.integer.default_number_of_names_per_group));
```
mohammedhemaid: How can I get access to this without context?

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -0,0 +1,69 @@
...
+        namesPerGroup.clearFocus();
+    }
+
+    private void applyNumberOfGroups() {
+        String numOfGroup = numGroups.getText().toString();
+        int inputtedNumber = numOfGroup.isEmpty() ? 0 :
+                Integer.parseInt(numGroups.getText().toString().trim());
+        if (inputtedNumber <= 0) {
+            numGroups.setText(R.string.default_number_of_groups);
+            settings.setNumOfGroups(context.getResources().getInteger(R.integer.default_number_of_groups));
```
mohammedhemaid: same here

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
Gear61: Context context = namesPerGroup.getContext() should work? Every view has a reference to its context.

Gear61 (approved):
LGTM (looks good to me), let's move onwards!

PR #94, opened 2020-01-12
[release] Made it easier to prepare for new releases

jchio001:
With this diff, releases only require 1 variable to be changed vs having to change version code and version number.


Gear61:
Sure, I guess. Tiny optimization is still an optimization.

PR #97, opened 2020-01-15
Adding custom speech to text

alaaalzibda:


app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+import com.joanzapata.iconify.widget.IconTextView;
+import com.randomappsinc.studentpicker.R;
+
+import java.util.ArrayList;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+import butterknife.OnClick;
+
+public class VoiceRecognizerDialogFragment extends DialogFragment implements RecognitionListener {
```
Gear61: We should use MaterialDialog instead. Its material design is backported and it abstracts a lot of the fragment management away. It also has a powerful set of APIs. We can just place a custom view inside of it and bind to it.

Also, we should create SpeechToTextManager and have it manage the UI (fragment) and speech recognition separately. We don't want to couple them together because it limits what we can do with them, makes the code less reusable, and makes our code prone to becoming monolithic (1 giant class). We can even wrap the MaterialDialog into its own component like SpeechToTextDialog and expose APIs to mutate the UI.

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public interface VoiceRecognizerInterface {
```
Gear61: nit: This is more in line with the rest of the app. I'm not a fan of putting the entity type (in this case "interface") into the name, since it's redundant information.
```suggestion
    public interface Listener {
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+        void spokenText(String spokenText);
```
Gear61: nit: More standard naming
```suggestion
        void onTextSpoken(String spokenText);
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @BindView(R.id.voice_icon) IconTextView voiceIcon;
```
Gear61: We aren't invoking any IconTextView specific APIs.
```suggestion
    @BindView(R.id.voice_icon) TextView voiceIcon;
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    @BindView(R.id.message) IconTextView message;
```
Gear61: Same as above.
```suggestion
    @BindView(R.id.message) TextView message;
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    @BindView(R.id.try_again) TextView tryAgain;
```
Gear61: We just need to turn this on/off right? View is the most generic we can go, so let's use that instead.
```suggestion
    @BindView(R.id.try_again) View tryAgain;
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private SpeechRecognizer mSpeechRecognizer;
```
Gear61: Let's be consistent with the rest of the app.
```suggestion
    private SpeechRecognizer speechRecognizer;
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    private Intent mSpeechRecognizerIntent;
```
Gear61: Same as above
```suggestion
    private Intent speechRecognizerIntent;
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    private Context context;
```
Gear61: We should be extremely careful storing contexts. Contexts are often activities, and if you don't clean up a class with a context properly, you might end up leaking an entire activity, which is a massive performance loss. We should avoid storing contexts at all if possible.

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    public VoiceRecognizerDialogFragment(Context context, VoiceRecognizerInterface signal) {
+        this.context = context;
+        this.signal = signal;
+    }
+
+    public VoiceRecognizerDialogFragment() {
+    }
+
+    @Override
+    public void onResume() {
```
Gear61: Be careful firing critical blocks in onResume(). It fires extremely often (instantiation, app background and return, return from activity on top, etc). In this case, this couples the speech recognition with the UI lifecycle, which makes the speech recognition kind of a side effect. With this design, the consumer of this module isn't really starting speech recognition, it's throwing up a fragment and the speech recognition just sort of happens. We should expose a more declarative API like startSpeechToText().

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    private void startListening() {
+        mSpeechRecognizer.setRecognitionListener(this);
+        mSpeechRecognizer.startListening(mSpeechRecognizerIntent);
+        changeUIStateToListening();
+    }
+
+    @Override
+    public void onDestroy() {
+        super.onDestroy();
+        if (mSpeechRecognizer != null) {
```
Gear61: We should err on the side of not having null checks; having too many null checks can suppress legitimate issues. If something can legitimately be null, mark it as @Nullable and explain when it can be null. In this case, it should always be non-null.

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+            mSpeechRecognizer.destroy();
+        }
+    }
+
+    @Override
+    public void onBeginningOfSpeech() {
+        message.setText(getString(R.string.waiting));
```
Gear61: nit: TextViews can take in a resource ID directly
```suggestion
        message.setText(R.string.waiting);
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+    public void onPartialResults(Bundle partialResults) {
+    }
+
+    @Override
+    public void onReadyForSpeech(Bundle params) {
+    }
+
+    @Override
+    public void onResults(Bundle results) {
+        ArrayList<String> matches = results.getStringArrayList(SpeechRecognizer.RESULTS_RECOGNITION);
```
Gear61: nit: We should always code to an interface when we can. A general principle is to "hide" underlying implementation as much as possible.
```suggestion
        List<String> matches = results.getStringArrayList(SpeechRecognizer.RESULTS_RECOGNITION);
```

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+        int i = 0;
+        String first = "";
+        for (String s : matches) {
+            if (i == 0) {
+                first = s;
+            }
+            i++;
+        }
+        signal.spokenText(first);
```
Gear61: I'm confused, can't we just grab the item at index 0 in the list? Just make sure the list isn't null or empty first.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -331,4 +357,11 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
         return super.onOptionsItemSelected(item);
     }
+
+    @Override
+    public void spokenText(String searchInput) {
+        if (searchInput != null) {
```
Gear61: The module should never pass back a null or empty string, because in that case, discernible text wasn't really spoken.

app/src/main/res/layout/voice_recognizble_layout.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.cardview.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
```
Gear61:
```suggestion
<androidx.cardview.widget.CardView
      xmlns:android="http://schemas.android.com/apk/res/android"
```

Gear61 (changes requested):
This is a great start! I'm surprised you got this working so quickly; this is not an easy feature to build.

You should avoid large pull requests, because it makes the code review really hard. To start, you will need to prove a LOT of things work in your demo, which will be hard on you and makes things easier to be missed. A lot of pieces are also dependent on one another, so if a piece at the base needs a change, you will need to change everything else. On the other hand, if we land a small chunk of the feature at a time, we can make sure the base is good before we start building on top of it, leading to less code changes. Please split up the pull request into smaller pieces, similar to what @mohammedhemaid is doing with the group maker. As mentioned in the issue, let's tackle just the permissions first.

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
alaaalzibda: So I have ta make SpeechToTextManager class that handles Voice Recognizing, also has showDialog function and of course using MaterialDialog with custom view.
Is that what you mean by saying "wrap the MaterialDialog into its own component" ?

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+import android.widget.TextView;
+import com.afollestad.materialdialogs.MaterialDialog;
+import com.joanzapata.iconify.widget.IconTextView;
+import com.randomappsinc.studentpicker.R;
+import java.util.List;
+import butterknife.BindView;
+import butterknife.ButterKnife;
+import butterknife.OnClick;
+
+public class SpeechToTextManager implements RecognitionListener {
```
Gear61, line 21:
```suggestion
public class SpeechToTextManager implements RecognitionListener {

```

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+    public interface Listener {
+        void onSpokenText(String spokenText);
```
Gear61:
```suggestion
        void onTextSpoken(String spokenText);
```

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @BindView(R.id.voice_icon) IconTextView voiceIcon;
```
Gear61: nit: Please code to an interface
```suggestion
    @BindView(R.id.voice_icon) TextView voiceIcon;
```

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+    private SpeechRecognizer speechRecognizer;
+    private Intent speechRecognizerIntent;
+    private Listener listener;
+    private MaterialDialog dialog;
+
+    public SpeechToTextManager(Context context, Listener listener) {
+        this.listener = listener;
+        speechRecognizer = SpeechRecognizer.createSpeechRecognizer(context);
+        speechRecognizerIntent = new Intent(RecognizerIntent.ACTION_RECOGNIZE_SPEECH);
+        speechRecognizerIntent.putExtra(RecognizerIntent.EXTRA_LANGUAGE_MODEL, RecognizerIntent.LANGUAGE_MODEL_WEB_SEARCH);
```
Gear61: We're not doing a web search, so this is probably better?
```suggestion
        speechRecognizerIntent.putExtra(RecognizerIntent.EXTRA_LANGUAGE_MODEL, RecognizerIntent.LANGUAGE_MODEL_FREE_FORM);
```

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+        speechRecognizerIntent.putExtra(RecognizerIntent.EXTRA_LANGUAGE, "en");
```
Gear61: This hardcodes the language to English right? Let's try pulling from Locale.getDefault() so the speech model matches the native language of the speaker (there are a lot of non-Americans using this app). I'm not sure if that value on its own though can be passed in here. I'll let you figure that out :)

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+        speechRecognizerIntent.putExtra(RecognizerIntent.EXTRA_PARTIAL_RESULTS, true);
```
Gear61: We don't use partial results. This is for features like Google Assistant showing your spoken text as you speak. We can do that in v2 of this feature.
```suggestion
        speechRecognizerIntent.putExtra(RecognizerIntent.EXTRA_PARTIAL_RESULTS, false);
```

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+        speechRecognizerIntent.putExtra(RecognizerIntent.EXTRA_CALLING_PACKAGE, context.getPackageName());
```
Gear61: Check this out: https://developer.android.com/reference/android/speech/RecognizerIntent.html#EXTRA_CALLING_PACKAGE

This flag is not meant to be used by developers. Let's delete this line. This is why we need to be careful with code copy-pasted from the internet; it's rarely optimal. Always check the code against the official documentation!

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+    public void onBufferReceived(byte[] buffer) {
+    }
+
+    @Override
+    public void onEndOfSpeech() {
+    }
+
+    @Override
+    public void onError(int error) {
+        if (error == 7) {
```
Gear61: Isn't there some constant in the SDK we can use? 7 is a magic number; we should avoid this since it makes the code hard to read.

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+    public void onResults(Bundle results) {
+        List<String> matches = results.getStringArrayList(SpeechRecognizer.RESULTS_RECOGNITION);
+        if (matches == null) {
+            return;
+        }
+        if (!matches.isEmpty()) {
+            listener.onSpokenText(matches.get(0));
+            dialog.dismiss();
+        } else {
+            changeUIStateToRetry();
```
Gear61, line 97: Should we run the retry logic for the null case as well? What's the difference? Can you dig into the documentation and find out?

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Override
+    public void onRmsChanged(float rmsdB) {
+    }
+
+    private void changeUIStateToListening() {
+        voiceIcon.setText(R.string.voice_icon_recognizer_white);
+        voiceIcon.setBackgroundResource(R.drawable.recognizable_voice_blue);
+        message.setText(R.string.list_name_input_speech_message);
+        tryAgain.setVisibility(View.INVISIBLE);
```
Gear61: Can we set this to View.GONE instead? We should avoid using INVISIBLE if possible since the view still takes up space and puts out a click hitbox, which can lead to weird experiences.

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -45,13 +45,15 @@
...
 
 public class MainActivity extends StandardActivity
-        implements NameListsAdapter.Delegate, RenameListDialog.Listener, DeleteListDialog.Listener {
+        implements NameListsAdapter.Delegate, RenameListDialog.Listener, DeleteListDialog.Listener, SpeechToTextManager.Listener {
 
     public static final String LIST_NAME_KEY = "listName";
 
     private static final int SPEECH_REQUEST_CODE = 1;
     private static final int IMPORT_FILE_REQUEST_CODE = 2;
     private static final int SAVE_IMPORT_REQUEST_CODE = 3;
```
Gear61, line 54: nit: Just to separate the different bundles of code in terms of purpose.
```suggestion
    private static final int SAVE_IMPORT_REQUEST_CODE = 3;
    
```

app/src/main/java/com/randomappsinc/studentpicker/home/MainActivity.java
```diff
@@ -331,4 +354,9 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
         return super.onOptionsItemSelected(item);
     }
+
+    @Override
+    public void onSpokenText(String searchInput) {
```
Gear61: nit: Can we move this under the voiceEntry() block to keep code working towards the same goal together? Better for organization.

app/src/main/res/drawable/recognizable_voice_blue.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
```
Gear61:
```suggestion
<shape
    xmlns:android="http://schemas.android.com/apk/res/android"
```

app/src/main/res/drawable/recognizable_voice_blue.xml
```diff
@@ -0,0 +1,5 @@
...
+    android:shape="oval" >
```
Gear61: nit: Better spacing and to avoid merge conflicts in the future
```suggestion
    android:shape="oval"
    >

```

app/src/main/res/drawable/recognizable_voice_error.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
```
Gear61:
```suggestion
<shape
    xmlns:android="http://schemas.android.com/apk/res/android"
```

app/src/main/res/drawable/recognizable_voice_error.xml
```diff
@@ -0,0 +1,11 @@
...
+    android:shape="oval"
+    >
+    <solid android:color="@android:color/white" />
```
Gear61: Using the Android white is probably pretty safe, but you never know. In general, it's better to be declarative and use your own resources instead of the Android SDK's. You never know what weird stuff can happen and Google developers can push bugs too. I remember Android O made the default accent color magenta or something, so a lot of apps started showing magenta everywhere.
```suggestion

    <solid android:color="color/white" />
```

app/src/main/res/values/strings.xml
```diff
@@ -70,4 +77,7 @@
     <string name="speech_unrecognized">
         We couldn\'t recognize what you said. Please try again.
     </string>
+    <string name="try_again">Try again</string>
+    <string name="waiting">{fa-cog spin 20sp}</string>
```
Gear61: nit: This is more descriptive.
```suggestion
    <string name="spinning_icon_while_waiting">{fa-cog spin 20sp}</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -70,4 +77,7 @@
...
+    <string name="did_not_catch">Didn\'t catch that, Try speaking again</string>
```
Gear61:
```suggestion
    <string name="did_not_catch_speech">We didn't catch what you said, please try again.</string>
```

app/src/main/res/values/styles.xml
```diff
@@ -76,4 +76,10 @@
         <item name="android:textSize">24sp</item>
         <item name="android:foreground">?attr/selectableItemBackground</item>
     </style>
+
+    <style name="SpeechDialogAnimation">
```
Gear61: nit: To be consistent with the rest of the codebase
```suggestion
    <style name="speech_dialog_animation">
```

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+        speechRecognizerIntent.putExtra(RecognizerIntent.EXTRA_CALLING_PACKAGE, context.getPackageName());
+        speechRecognizer.setRecognitionListener(this);
+
+        dialog = new MaterialDialog.Builder(context)
+                .customView(R.layout.voice_recognizble_layout, false)
+                .build();
+        ButterKnife.bind(this, dialog);
+    }
+
+    public void showSpeechDialog() {
```
Gear61: Speech to text having UI is another thing we should hide from the caller. The caller should just know if I use this module and start the flow, I will eventually get back a string (assuming all goes well). It should be complete magic; the caller shouldn't even know there's a dialog somewhere in the mix. This API is also inaccurate, because it doesn't just show a dialog, it starts listening to the user, making the 2nd line feel like more of a side effect. Let's change this API to the suggestion below. Also, we still need to take in a customizable message resource, similar to what our implementation did before.
```suggestion
    public void startSpeechToTextFlow(@StringRes int messageId) {
```

app/src/main/res/layout/voice_recognizble_layout.xml
```diff
@@ -0,0 +1,48 @@
...
+    android:layout_height="wrap_content"
+    app:cardCornerRadius="8dp"
+    >
+
+    <LinearLayout
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:layout_margin="32dp"
+        android:gravity="center"
```
Gear61: Since the height of the container is wrap_content, center is excessive.
```suggestion
        android:gravity="center_horizontal"
```

app/src/main/res/layout/voice_recognizble_layout.xml
```diff
@@ -0,0 +1,48 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="16dp"
+            android:background="@drawable/add_item_border"
+            android:paddingStart="16dp"
+            android:paddingTop="8dp"
+            android:paddingEnd="16dp"
+            android:paddingBottom="8dp"
+            android:text="@string/try_again"
+            android:textColor="@color/blue_200"
```
Gear61, line 48: We should explicitly code the textSize and not leave it to the SDK default. 19sp should be good?

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
```diff
@@ -0,0 +1,127 @@
...
+public class SpeechToTextManager implements RecognitionListener {
+    public interface Listener {
+        void onSpokenText(String spokenText);
+    }
+
+    @BindView(R.id.voice_icon) IconTextView voiceIcon;
+    @BindView(R.id.message) TextView message;
+    @BindView(R.id.try_again) View tryAgain;
+
+    private SpeechRecognizer speechRecognizer;
```
Gear61, line 30: Don't we need to release/destroy this resource somewhere? We can expose a cleanUp() API in this class that the fragment/activity calls in their onDestroy() block. Otherwise, I think we might leak something?

app/src/main/res/layout/voice_recognizble_layout.xml
```diff
@@ -0,0 +1,48 @@
...
+            android:layout_height="wrap_content"
+            android:padding="28dp"
+            />
+
+        <com.joanzapata.iconify.widget.IconTextView
+            android:id="@+id/message"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginTop="16dp"
+            android:textColor="@color/dark_gray"
```
Gear61, line 33: This also needs a textSize. 17sp? I'm just guessing here. Keep playing around with it until it looks good.

app/src/main/res/layout/voice_recognizble_layout.xml
```diff
@@ -0,0 +1,48 @@
...
+        android:layout_margin="32dp"
+        android:gravity="center"
+        android:orientation="vertical"
+        >
+
+        <com.joanzapata.iconify.widget.IconTextView
+            android:id="@+id/voice_icon"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:padding="28dp"
```
Gear61, line 23: Given that this is an icon, the textSize needs to be bigger. So 32dp? 40dp?

Gear61 (changes requested):
This is way better than before; keep it up!

app/src/main/java/com/randomappsinc/studentpicker/common/VoiceRecognizerDialogFragment.java
Gear61: Wrapping the MaterialDialog in its own component is creating yet another class just for the dialog UI. That's kind of a bonus; you don't need to do it.

Gear61:
Also, can you attach a screen record of the feature in action? Thanks :)

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
alaaalzibda: Locale.getDefault().getLanguage() returns language code.

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
alaaalzibda: It's because I used wrap content for the dialog height not a fixed height, changing the visibility to GONE will make the dialog change the height.
I try to avoid set hard coded width and height for any component.

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
alaaalzibda: It could be null or empty because the recognizer couldn't get the results correctly for some reasons.
 So I think yeah, we should retry the speech recognition in this case.

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
alaaalzibda: yeah, we should use SpeechRecognizer.ERROR_NO_MATCH which is 7

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
alaaalzibda: The message we send to startSpeechToTextFlow is the one which is gonna be shown on the dialog?
If yes then I think it's better to make changeUIStateToListening, changeUIStateToRetry to only one function changeUIStates, then send the variables for it to make it perform as we wish!

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
Gear61: Yeah, that's what I thought, it returns a Locale right? I think there's some API Locale exposes that we can use.

app/src/main/res/layout/voice_recognizble_layout.xml
alaaalzibda: I've set the size on strings file
{ion-android-microphone 40sp #FFFFFF}

app/src/main/java/com/randomappsinc/studentpicker/common/SpeechToTextManager.java
alaaalzibda: startSpeechToTextFlow API should have 3 purposes:
1- Showing SpeechToTextDialog
2- Start listening
3- Changing UI states to listening

alaaalzibda:
A video shows custom speech to text.
[Speech to text video](https://drive.google.com/open?id=1Ch44NInn1rBnREC0hAfLiqYfLxgygEc6)

Gear61 (approved):
Looks good! Merging this in.

PR #105, opened 2020-01-20
Updating rename dialog to be more usable

alaaalzibda:
Screenshot_20200120-024616

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
@@ -19,7 +19,6 @@ public class DataSource {
     private SQLiteDatabase database;
     private MySQLiteHelper dbHelper;
 
-    // Constructor
     public DataSource(Context context) {
         dbHelper = new MySQLiteHelper(context);
     }
@@ -189,7 +188,7 @@ public class DataSource {
         return nameAmounts;
     }
 
-    public void renamePeople(String oldName, String newName, String listName, int amount) {
+    void renamePeople(String oldName, String newName, String listName, int amount) {
         removeNames(oldName, listName, amount);
         addNames(newName, listName, amount);
     }


app/src/main/java/com/randomappsinc/studentpicker/editing/RenameDialog.java
```diff
@@ -37,10 +36,12 @@ public RenameDialog(Context context, final Listener listener) {
                 .alwaysCallInputCallback()
                 .negativeText(android.R.string.no)
                 .positiveText(R.string.next)
+                .onNeutral((dialog, which) -> {
+                    dialog.getInputEditText().setText("" + currentMaxAmount);
```
Gear61: This feels kind of hacky, and I think it's slower than just converting it normally.
```suggestion
                    dialog.getInputEditText().setText(String.valueOf(currentMaxAmount));
```

app/src/main/java/com/randomappsinc/studentpicker/editing/RenameDialog.java
```diff
@@ -64,14 +65,22 @@ public RenameDialog(Context context, final Listener listener) {
                 .build();
     }
 
-    public void startRenamingProcess(String name, int maxAmount) {
+    private void handlePositiveAmountToRename() {
```
Gear61: This isn't worth the extra method. Especially in the neutral case where you already know exactly the amount to rename (it's just currentMaxAmount) and are incurring a performance loss from the integer parsing and string trimming.

app/src/main/java/com/randomappsinc/studentpicker/editing/RenameDialog.java
```diff
@@ -64,14 +65,22 @@ public RenameDialog(Context context, final Listener listener) {
...
         currentName = name;
         currentMaxAmount = maxAmount;
         renamingDialog.getInputEditText().setText(name);
         if (maxAmount > 1) {
+            renameAmountDialog.setActionButton(DialogAction.NEUTRAL, R.string.all_of_them);
+            renameAmountDialog.setContent(R.string.multiple_renames_title, name , maxAmount);
             EditText input = renameAmountDialog.getInputEditText();
             if (input != null) {
-                input.setText("");
+                input.setText("" + maxAmount);
```
Gear61:
```suggestion
                input.setText("" + String.valueOf(maxAmount));
```

app/src/main/res/values/edit_name_list_strings.xml
```diff
@@ -39,7 +39,8 @@
     <string name="multiple_deletions_title">How many copies of this name do you want to delete?</string>
     <string name="names_deleted">Name(s) deleted.</string>
 
-    <string name="multiple_renames_title">How many copies of this name do you want to change?</string>
+    <string name="multiple_renames_title">How many copies of \'%s\' do you want to change? You currently have %s copies of this name.</string>
```
Gear61: Let's split this up, so it doesn't flood people's IDEs. Also, we should use %d for numbers. Lastly, we should use double quotes, not single quotes. Double quotes are kind of wonky in XML though, so you might need to just use %s and then add the quotes in Java.
```suggestion
    <string name="multiple_renames_title">
        How many copies of \'%s\' do you want to change?
        You currently have %d copies of this name.
    </string>
```

Gear61 (changes requested):
Thanks for doing this and attaching the screenshot!

app/src/main/java/com/randomappsinc/studentpicker/editing/RenameDialog.java
alaaalzibda: I put String.ValuveOf at first but then I looked at it and said "" + will make code shorter

Gear61 (changes requested):
Sorry, but I'm putting my foot down here; having everything together makes the code hard to review and isn't teaching good habits. Please split these up into separate pull requests or merge your commits in 1 at a time. This will be a good opportunity for you to learn about managing commit history and branching. Here's one way you can split things up without losing any code:

1. Create a branch called "all-my-changes".
2. Go back to your master branch. Use git reset --hard HEAD ~X, where X is the amount of commits you want to delete. In this case, rewind back to the point where you only have rename dialog commits
3. Update your master which should update this pull request
4. After the pull request gets merged, you can merge "all-my-changes" into your master again and rewind the commits as you want.

Another way is to create a bunch of new branches (so 4 in this case?) and cherry pick each commit onto each branch, so each branch has its own issue. And then submit a bunch of pull requests for each change. This is how you work on things in parallel versus the sequential style of having things merged in 1 at a time. This will probably lead to more merge conflicts on your end though, since you have to keep updating against master.

app/src/main/res/values/edit_name_list_strings.xml
```diff
@@ -39,7 +39,11 @@
     <string name="multiple_deletions_title">How many copies of this name do you want to delete?</string>
     <string name="names_deleted">Name(s) deleted.</string>
 
-    <string name="multiple_renames_title">How many copies of this name do you want to change?</string>
+    <string name="multiple_renames_title">
+        How many copies of %s do you want to change?
+         You currently have %d copies of this name.
```
Gear61: nit: Newline in XML will be interpreted as a blank space in rendering IIRC. Removing this should prevent double spacing.
```suggestion
        You currently have %d copies of this name.
```

Gear61 (changes requested):
One more tiny change to fix a slight UI bug, but everything else looks good!

PR #112, opened 2020-01-23
Make name choosing history manager

alaaalzibda:
Name choosing history

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingHistoryManager.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingHistoryManager.java
@@ -1,0 +1,55 @@
+package com.randomappsinc.studentpicker.choosing;
+
+import android.content.Context;
+
+import androidx.annotation.NonNull;
+
+import com.afollestad.materialdialogs.MaterialDialog;
+import com.randomappsinc.studentpicker.R;
+import com.randomappsinc.studentpicker.models.ListInfo;
+import com.randomappsinc.studentpicker.utils.NameUtils;
+import com.randomappsinc.studentpicker.utils.UIUtils;
+
+public class NameChoosingHistoryManager {
+
+    public interface Listener {
+        void onEmptyHistoryNames();
+    }
+
+    private ListInfo listInfo;
+    private String namesHistory;
+    private MaterialDialog dialog;
+    private Listener listener;
+
+    NameChoosingHistoryManager(ListInfo listInfo, Context context, @NonNull Listener listener) {
+        this.listInfo = listInfo;
+        this.listener = listener;
+
+        dialog = new MaterialDialog.Builder(context)
+                .title(R.string.chosen_names_history)
+                .positiveText(android.R.string.yes)
+                .neutralText(R.string.clear)
+                .negativeText(R.string.copy_text)
+                .onNeutral((dialog, which) -> {
+                    listInfo.clearNameHistory();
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingHistoryManager.java
```diff
@@ -0,0 +1,55 @@
...
+import com.randomappsinc.studentpicker.utils.UIUtils;
+
+public class NameChoosingHistoryManager {
+
+    public interface Listener {
+        void onEmptyHistoryNames();
+    }
+
+    private ListInfo listInfo;
+    private String namesHistory;
```
Gear61: This member variable isn't necessary, since the names history is always changing and we always just grab it on demand and immediately render it (local usage).

app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingHistoryManager.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.afollestad.materialdialogs.MaterialDialog;
+import com.randomappsinc.studentpicker.R;
+import com.randomappsinc.studentpicker.models.ListInfo;
+import com.randomappsinc.studentpicker.utils.NameUtils;
+import com.randomappsinc.studentpicker.utils.UIUtils;
+
+public class NameChoosingHistoryManager {
+
+    public interface Listener {
```
Gear61: Setting up a listener just to show a Toast is sort of underwhelming, especially when this class already shows a Toast in another use case. To minimize logic in the fragment, let's just show the Toast in this class. The dialog always has access to a context since it's a fragment with views.

Gear61 (changes requested):

app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingHistoryManager.java
```diff
@@ -0,0 +1,46 @@
...
+                .onNegative((dialog, which) -> NameUtils.copyNamesToClipboard(
+                        listInfo.getNameHistoryFormatted(),
+                        null,
+                        0,
+                        true,
+                        dialog.getContext()))
+                .build();
+    }
+
+    void showNamesHistory() {
```
Gear61: nit: Since it might not happen due to empty case
```suggestion
    void maybeShowNamesHistory() {
```

app/src/main/java/com/randomappsinc/studentpicker/choosing/NameChoosingHistoryManager.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private ListInfo listInfo;
+    private MaterialDialog dialog;
+
+    NameChoosingHistoryManager(ListInfo listInfo, Context context) {
+        this.listInfo = listInfo;
+
+        dialog = new MaterialDialog.Builder(context)
+                .title(R.string.chosen_names_history)
+                .positiveText(android.R.string.yes)
```
Gear61: I think there's an R.string.okay we can use. Otherwise, let's define it in the common strings.xml file and use it. Using the Android defaults isn't super readable IMHO (yes is actually "Ok") and it leaves us vulnerable to SDK changes.

Gear61 (changes requested):
Almost there!

Gear61 (approved):
Awesome, thanks for the refactoring!

PR #113, opened 2020-01-23
Name duplication feature

alaaalzibda:
[Video for name duplication feature](https://drive.google.com/open?id=1uyz1lszd1OQFYHhJHltRVKXlLn2uoHTM)

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
@@ -23,9 +23,7 @@ public class NameListDataManager {
     }
 
     public interface AdjustAmountListener {
-        void onNamesRemoved(String name, int amountToDelete);
-
-        void onNamesAdded(String name, int amountToAdd);
+        void onAdjustNameAmount(String name, int newAmount);
     }
 
     private static NameListDataManager instance;
@@ -75,16 +73,16 @@ public class NameListDataManager {
         }
     }
 
-    public void setNameAmount(String name, int amountToHave, int currentAmount, Context context, String listName) {
-        if (currentAmount > amountToHave) {
-            int amountToDelete = currentAmount - amountToHave;
-            deleteName(context, name, amountToDelete, listName);
-            adjustAmountListener.onNamesRemoved(name, amountToDelete);
+    public void setNameAmount(String name, int newAmount, int currentAmount, Context context, String listName) {
+        DataSource dataSource = new DataSource(context);
+        if (currentAmount > newAmount) {
+            int amountToDelete = currentAmount - newAmount;
+            dataSource.removeNames(name, listName, amountToDelete);
         } else {
-            int amountToAdd = amountToHave - currentAmount;
-            addName(context, name, amountToAdd, listName);
-            adjustAmountListener.onNamesAdded(name, amountToAdd);
+            int amountToAdd = newAmount - currentAmount;
+            dataSource.addNames(name, listName, amountToAdd);
         }
+        adjustAmountListener.onAdjustNameAmount(name, newAmount);
     }
 
     public void changeName(Context context, String oldName, String newName, int amount, String listName) {
--- a/app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
@@ -22,7 +22,14 @@ public class NameListDataManager {
         void onNameListsImported(Map<String, Integer> nameAmounts, String listName);
     }
 
+    public interface AdjustAmountListener {
+        void onNamesRemoved(String name, int amountToDelete);
+
+        void onNamesAdded(String name, int amountToAdd);
+    }
+
     private static NameListDataManager instance;
+    private AdjustAmountListener adjustAmountListener;
 
     public static NameListDataManager get() {
         if (instance == null) {
@@ -48,6 +55,10 @@ public class NameListDataManager {
         listeners.remove(listener);
     }
 
+    public void registerAmountListener(AdjustAmountListener adjustAmountListener) {
+        this.adjustAmountListener = adjustAmountListener;
+    }
+
     public void addName(Context context, String name, int amount, String listName) {
         DataSource dataSource = new DataSource(context);
         dataSource.addNames(name, listName, amount);
@@ -61,6 +72,18 @@ public class NameListDataManager {
         dataSource.removeNames(name, listName, amount);
         for (Listener listener : listeners) {
             listener.onNameDeleted(name, amount, listName);
+        }
+    }
+
+    public void setNameAmount(String name, int amountToHave, int currentAmount, Context context, String listName) {
+        if (currentAmount > amountToHave) {
+            int amountToDelete = currentAmount - amountToHave;
+            deleteName(context, name, amountToDelete, listName);
+            adjustAmountListener.onNamesRemoved(name, amountToDelete);
+        } else {
+            int amountToAdd = amountToHave - currentAmount;
+            addName(context, name, amountToAdd, listName);
+            adjustAmountListener.onNamesAdded(name, amountToAdd);
         }
     }
 
--- a/app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
@@ -22,12 +22,7 @@ public class NameListDataManager {
         void onNameListsImported(Map<String, Integer> nameAmounts, String listName);
     }
 
-    public interface AdjustAmountListener {
-        void onAdjustNameAmount(String name, int newAmount);
-    }
-
     private static NameListDataManager instance;
-    private AdjustAmountListener adjustAmountListener;
 
     public static NameListDataManager get() {
         if (instance == null) {
@@ -53,10 +48,6 @@ public class NameListDataManager {
         listeners.remove(listener);
     }
 
-    public void registerAmountListener(AdjustAmountListener adjustAmountListener) {
-        this.adjustAmountListener = adjustAmountListener;
-    }
-
     public void addName(Context context, String name, int amount, String listName) {
         DataSource dataSource = new DataSource(context);
         dataSource.addNames(name, listName, amount);
@@ -74,15 +65,13 @@ public class NameListDataManager {
     }
 
     public void setNameAmount(String name, int newAmount, int currentAmount, Context context, String listName) {
-        DataSource dataSource = new DataSource(context);
         if (currentAmount > newAmount) {
             int amountToDelete = currentAmount - newAmount;
-            dataSource.removeNames(name, listName, amountToDelete);
+            deleteName(context, name, amountToDelete, listName);
         } else {
             int amountToAdd = newAmount - currentAmount;
-            dataSource.addNames(name, listName, amountToAdd);
+            addName(context, name, amountToAdd, listName);
         }
-        adjustAmountListener.onAdjustNameAmount(name, newAmount);
     }
 
     public void changeName(Context context, String oldName, String newName, int amount, String listName) {
--- a/app/src/main/java/com/randomappsinc/studentpicker/editing/DuplicationDialog.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/editing/DuplicationDialog.java
@@ -11,35 +11,41 @@ import com.randomappsinc.studentpicker.R;
 public class DuplicationDialog {
 
     public interface Listener {
-        void onDuplicationSubmitted(String name, int amountToAdd);
+        void onDuplicationSubmitted(String name, int amountToHave, int currentAmount);
     }
 
     private MaterialDialog dialog;
     private String currentName;
+    private int currentAmount;
 
     public DuplicationDialog(Context context, final Listener listener) {
         dialog = new MaterialDialog.Builder(context)
                 .content(R.string.cloning_title)
                 .inputType(InputType.TYPE_CLASS_NUMBER)
                 .input(context.getString(R.string.num_copies), "", (dialog, input) -> {
-                    boolean isValid = input.length() > 0 && (Integer.parseInt(input.toString()) > 0);
-                    dialog.getActionButton(DialogAction.POSITIVE).setEnabled(isValid);
+                    if (input.length() > 0) {
+                        int numCopies = Integer.parseInt(input.toString());
+                        boolean isValid = numCopies > 0 && numCopies != currentAmount;
+                        dialog.getActionButton(DialogAction.POSITIVE).setEnabled(isValid);
+                    }
                 })
                 .alwaysCallInputCallback()
                 .negativeText(android.R.string.no)
-                .positiveText(R.string.add)
+                .positiveText(R.string.submit)
                 .onPositive((dialog, which) -> {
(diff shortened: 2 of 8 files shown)


app/src/main/res/values/edit_name_list_strings.xml
```diff
@@ -16,10 +16,19 @@
     <string name="list_icon">{ion-android-list}</string>
     <string name="rename_person">Rename \"%s\"</string>
     <string name="delete_name">Delete \"%s\"</string>
+    <string name="duplicates_deleted">%d Duplicate(s) deleted</string>
```
Gear61:
```suggestion
    <string name="name_amount_adjusted">Name amount adjusted.</string>
```

app/src/main/res/values/edit_name_list_strings.xml
```diff
@@ -16,10 +16,19 @@
...
+    <string name="adjust_name_amount">Adjust name amount</string>
+
     <string name="duplicate">Duplicate \"%s\"</string>
-    <string name="cloning_title">How many copies of this name do you want to add?</string>
+
+    <string name="cloning_title">
```
Gear61: This isn't the title of the dialog, it's the content.
```suggestion
    <string name="name_adjustment_message">
```

app/src/main/res/values/edit_name_list_strings.xml
```diff
@@ -16,10 +16,19 @@
...
+         How many copies of %s do you want to have?
+         You currently have %d copies of this name.
+     </string>
```
Gear61: nit: Align with the opening tag above.
```suggestion
    </string>
```

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -159,21 +159,22 @@ void addNameWithVoice() {
...
+        deleteNameDialog.startDeletionProcess(name, amountOfEditChoiceFlow(name));
     }
 
     @Override
     public void onDuplicationChosen(String name) {
-        duplicationDialog.show(name);
+        duplicationDialog.show(name, amountOfEditChoiceFlow(name));
+    }
+
+    private int amountOfEditChoiceFlow(String name) {
```
Gear61: Adding a method always comes with a cost because you want to stay underneath 65,536 methods to avoid multi-dexing, it adds latency, and it can make the code harder to read. I don't think this is worth the extra method, and the name is confusing (it has "edit" in it but is used for the "deletion" and "duplication" flow). We should continue inlining this code, and we can improve on my prior code by just merging the 2 lines together. Assigning the listinfo to a variable is kind of overkill.

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -159,21 +159,22 @@ void addNameWithVoice() {
...
     @Override
     public void onDeleteChosen(String name) {
-        ListInfo listInfo = namesAdapter.getListInfo();
-        int currentAmount = listInfo.getInstancesOfName(name);
-        deleteNameDialog.startDeletionProcess(name, currentAmount);
+        deleteNameDialog.startDeletionProcess(name, amountOfEditChoiceFlow(name));
     }
 
     @Override
     public void onDuplicationChosen(String name) {
```
Gear61: We should try to keep the user experience and the implementation as harmonized as possible so our code is easy to read and navigate. Let's change all the duplication terminology. So we should change this to "onNameAdjustmentChosen".

app/src/main/java/com/randomappsinc/studentpicker/editing/DuplicationDialog.java
```diff
@@ -11,35 +11,41 @@
...
+    private int currentAmount;
 
     public DuplicationDialog(Context context, final Listener listener) {
         dialog = new MaterialDialog.Builder(context)
                 .content(R.string.cloning_title)
                 .inputType(InputType.TYPE_CLASS_NUMBER)
                 .input(context.getString(R.string.num_copies), "", (dialog, input) -> {
-                    boolean isValid = input.length() > 0 && (Integer.parseInt(input.toString()) > 0);
-                    dialog.getActionButton(DialogAction.POSITIVE).setEnabled(isValid);
+                    if (input.length() > 0) {
```
Gear61: I think this code breaks the case where you type in a valid input and then delete it back to nothing (the button will stay enabled). If input.length() is 0, we aren't mutating the dialog's positive action button.

app/src/main/java/com/randomappsinc/studentpicker/editing/DuplicationDialog.java
```diff
@@ -11,35 +11,41 @@
...
+                        int numCopies = Integer.parseInt(input.toString());
+                        boolean isValid = numCopies > 0 && numCopies != currentAmount;
+                        dialog.getActionButton(DialogAction.POSITIVE).setEnabled(isValid);
```
Gear61: I don't think we should limit the user even though it's a weird case. If the user sets the amount to 0, just let them delete all the names.

app/src/main/java/com/randomappsinc/studentpicker/editing/DuplicationDialog.java
```diff
@@ -11,35 +11,41 @@
 public class DuplicationDialog {
 
     public interface Listener {
-        void onDuplicationSubmitted(String name, int amountToAdd);
+        void onDuplicationSubmitted(String name, int amountToHave, int currentAmount);
```
Gear61:
```suggestion
        void onNameAmountAdjustmentSubmitted(String name, int newAmount, int currentAmount);
```

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -195,10 +196,18 @@ public void onDeletionSubmitted(String name, int amountToDelete) {
...
 
     @Override
-    public void onDuplicationSubmitted(String name, int amountToAdd) {
-        nameListDataManager.addName(getContext(), name, amountToAdd, listName);
-        namesAdapter.addNames(name, amountToAdd);
-        UIUtils.showSnackbar(parent, R.string.clones_added);
+    public void onDuplicationSubmitted(String name, int amountToHave, int currentAmount) {
+        if (currentAmount > amountToHave) {
+            int amountToDelete = currentAmount - amountToHave;
+            nameListDataManager.deleteName(getContext(), name, amountToDelete, listName);
```
Gear61: Let's move this logic into NameListDataManager to keep the fragment slim. We can expose a setNameAmount() API.

Gear61 (changes requested):
This is a good start! This issue is harder than it seems on the surface since it requires a lot of architectural changes, but it will be a good exercise in writing clean code.

app/src/main/java/com/randomappsinc/studentpicker/editing/DuplicationDialog.java
alaaalzibda: You're right, I didn't notice that case

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -41,7 +42,7 @@
 
 public class EditNameListFragment extends Fragment implements
         NameEditChoicesDialog.Listener, RenameDialog.Listener, DeleteNameDialog.Listener,
-        DuplicationDialog.Listener, MergeNameListsDialog.Listener, SpeechToTextManager.Listener {
+        DuplicationDialog.Listener, MergeNameListsDialog.Listener, SpeechToTextManager.Listener, NameListDataManager.AdjustAmountListener {
```
Gear61: We should rename DuplicationDialog to NameAmountAdjustmentDialog.

app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
```diff
@@ -64,6 +75,18 @@ public void deleteName(Context context, String name, int amount, String listName
         }
     }
 
+    public void setNameAmount(String name, int amountToHave, int currentAmount, Context context, String listName) {
```
Gear61:
```suggestion
    public void setNameAmount(String name, int newAmount, int currentAmount, Context context, String listName) {
```

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -195,10 +196,8 @@ public void onDeletionSubmitted(String name, int amountToDelete) {
     }
 
     @Override
-    public void onDuplicationSubmitted(String name, int amountToAdd) {
-        nameListDataManager.addName(getContext(), name, amountToAdd, listName);
-        namesAdapter.addNames(name, amountToAdd);
-        UIUtils.showSnackbar(parent, R.string.clones_added);
+    public void onNameAmountAdjustmentSubmitted(String name, int amountToHave, int currentAmount) {
+       nameListDataManager.setNameAmount(name, amountToHave, currentAmount, getActivity(), listName);
```
Gear61: I think we can just show a Snackbar here instead of invoking the NameListDataManager and then having it communicate back to us. Also, similar to how we exposed a new API in NameListDataManager, we should just expose a new API in namesAdapter as well. We can also clean up the old purely additive APIs along the way as well.

Gear61 (changes requested):
I like the progress! Let's keep adding more powerful APIs instead of trying to "hack around" the limitations we have with our current API.

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -253,4 +251,10 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
         return super.onOptionsItemSelected(item);
     }
+
+    @Override
+    public void onAdjustNameAmount(String name, int newAmount) {
```
Gear61: We should just do this in onNameAmountAdjustmentSubmitted, since it will happen immediately after it. It's not worth adding another listener.

app/src/main/java/com/randomappsinc/studentpicker/database/NameListDataManager.java
```diff
@@ -64,6 +73,18 @@ public void deleteName(Context context, String name, int amount, String listName
...
+    public void setNameAmount(String name, int newAmount, int currentAmount, Context context, String listName) {
+        DataSource dataSource = new DataSource(context);
+        if (currentAmount > newAmount) {
+            int amountToDelete = currentAmount - newAmount;
+            dataSource.removeNames(name, listName, amountToDelete);
+        } else {
+            int amountToAdd = newAmount - currentAmount;
+            dataSource.addNames(name, listName, amountToAdd);
+        }
+        adjustAmountListener.onAdjustNameAmount(name, newAmount);
```
Gear61: We need to invoke onNameAdded() and onNameDeleted() in all the listeners. This is because changes to the edit list page need to be reflected in the name choice page (imagine deleting a name and then being able to choose it). This is the fastest and best experience for the user but leads to complexity for us unfortunately.

app/src/main/java/com/randomappsinc/studentpicker/editing/NameAmountAdjustmentDialog.java
```diff
@@ -8,38 +8,41 @@
...
+                .positiveText(R.string.submit)
                 .onPositive((dialog, which) -> {
                     int numCopies = Integer.parseInt(dialog.getInputEditText().getText().toString().trim());
-                    listener.onDuplicationSubmitted(currentName, numCopies);
+                    listener.onNameAmountAdjustmentSubmitted(currentName, numCopies, currentAmount);
                 })
                 .build();
         // Cap the amount you can make at 999
-        dialog.getInputEditText().setFilters(new InputFilter[] {new InputFilter.LengthFilter(3)});
+        dialog.getInputEditText().setFilters(new InputFilter[]{new InputFilter.LengthFilter(3)});
```
Gear61:
```suggestion
        dialog.getInputEditText().setFilters(new InputFilter[] {new InputFilter.LengthFilter(3)});
```

app/src/main/java/com/randomappsinc/studentpicker/editing/NameAmountAdjustmentDialog.java
```diff
@@ -8,38 +8,41 @@
...
 
-    public DuplicationDialog(Context context, final Listener listener) {
+    public NameAmountAdjustmentDialog(Context context, final Listener listener) {
         dialog = new MaterialDialog.Builder(context)
-                .content(R.string.cloning_title)
+                .content(R.string.name_adjustment_message)
                 .inputType(InputType.TYPE_CLASS_NUMBER)
                 .input(context.getString(R.string.num_copies), "", (dialog, input) -> {
-                    boolean isValid = input.length() > 0 && (Integer.parseInt(input.toString()) > 0);
+                    boolean isValid = input.length() > 0 && Integer.parseInt(input.toString()) != currentAmount;
```
Gear61, line 26: Negative numbers aren't possible right? Just want to make sure.

app/src/main/java/com/randomappsinc/studentpicker/models/ListInfo.java
```diff
@@ -90,6 +90,19 @@ public void removeNames(String name, int amount) {
         }
     }
 
+    public void setNameAmount(String name, int newAmount) {
+        if (nameAmounts.containsKey(name)) {
+            int currentAmount = nameAmounts.get(name);
+            if (currentAmount > newAmount) {
```
Gear61: At this level, we shouldn't invoke other APIs since that will lead to duplicated code. We can just update the HashMap directly. Make sure to keep numInstances updated properly though.

Gear61 (changes requested):

app/src/main/java/com/randomappsinc/studentpicker/models/ListInfo.java
```diff
@@ -90,6 +90,18 @@ public void removeNames(String name, int amount) {
...
 
+    public void setNameAmount(String name, int newAmount) {
+        if (nameAmounts.containsKey(name)) {
+            if (newAmount != 0) {
+                nameAmounts.put(name, newAmount);
+            } else {
+                nameAmounts.remove(name);
+                names.remove(name);
+            }
+            numInstances = newAmount;
```
Gear61: numInstances is the total number of names in the list, so I think this change will break the number of names indicator at the top. We need to change it by the delta we incurred on the name (either add or remove). We will need to get the current number of instances of this name from the HashMap to compute this.

Gear61 (changes requested):

alaaalzibda:
[Video shows test cases for setNameAmount API](https://drive.google.com/open?id=1ksgqDv9C3xogrtkXGS3g5JFKXL1t6P6C)

app/src/main/java/com/randomappsinc/studentpicker/models/ListInfo.java
```diff
@@ -90,6 +90,23 @@ public void removeNames(String name, int amount) {
         }
     }
 
+    public void setNameAmount(String name, int newAmount) {
+        if (nameAmounts.containsKey(name)) {
+            int currentAmount = nameAmounts.get(name);
```
Gear61, line 95: You got it!

Gear61 (approved):

Gear61:
@ALAA999, there's some merge conflicts, can you resolve them? Thanks!

PR #114, opened 2020-01-23
Update EditNameList to use RecyclerView with SimpleDividerItemDecoration

alaaalzibda:
switch to recyclerview

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListAdapter.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListAdapter.java
@@ -4,8 +4,10 @@ import android.content.Context;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.BaseAdapter;
 import android.widget.TextView;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
 
 import com.randomappsinc.studentpicker.R;
 import com.randomappsinc.studentpicker.database.DataSource;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListAdapter.java
```diff
@@ -75,46 +84,49 @@ public void importNamesFromList(ListInfo updatedListInfo) {
...
+    @Override
+    public int getItemCount() {
+        return content.getNumNames();
+    }
+
+    public String getItem(int position) {
+        return content.getName(position);
+    }
+
+    class NameViewHolder extends RecyclerView.ViewHolder implements View.OnClickListener{
```
Gear61, line 115: Instead of doing this implements thing, let's stay consistent by continuing the leverage ButterKnife. Give the parent view of the cell an ID like "parent" and use @OnClick(R.id.parent).

app/src/main/res/layout/lists_with_add_content.xml
```diff
@@ -78,10 +79,12 @@
             style="@style/num_names"
             />
 
-        <ListView
+        <androidx.recyclerview.widget.RecyclerView
             android:id="@+id/content_list"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
+            android:clipToPadding="false"
```
Gear61: We don't need clipToPadding="false", since there's no FloatingActionButton at the bottom.

app/src/main/java/com/randomappsinc/studentpicker/editing/EditNameListFragment.java
```diff
@@ -127,8 +130,8 @@ void addItem() {
         }
     }
 
-    @OnItemClick(R.id.content_list)
-    void showNameOptions(final int position) {
+    @Override
+    public void showNameOptions(final int position) {
         nameEditChoicesDialog.showChoices(namesAdapter.getItem(position));
```
Gear61: Hmmm, now that I think about this, having the adapter fire this callback and then have the fragment read from the adapter seems kind of roundabout. Let's just pass the name here directly.

Gear61 (changes requested):
This is a good start!

Gear61 (approved):
Awesome, looks great to me!

PR #116, opened 2020-01-28
Remove persistence of grouping settings

mohammedhemaid:
The video after the removing:
https://photos.app.goo.gl/JJxTnYz54jWKJPCg9
sorry for making the first letter of the commit in lowercase

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
@@ -1,6 +1,5 @@
 package com.randomappsinc.studentpicker.grouping;
 
-import android.content.Context;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.Menu;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
```diff
@@ -48,7 +48,6 @@ public static GroupingMakingFragment getInstance(String listName) {
     private String listName;
     private DataSource dataSource;
```
Gear61: We can remove DataSource too right?

Gear61 (changes requested):

Gear61:
Commit casing isn't a huge deal, don't worry about it. :)

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
mohammedhemaid: @Gear61  you mean to make it as a local variable?

mohammedhemaid: or making an object of it like this:
image

Gear61 (approved):
Awesome, looks good to me! Merging.

PR #118, opened 2020-02-02
Set up the adapter with the cards

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/s6sU4KLddKTmw4aAA

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
@@ -8,28 +8,31 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.TextView;
-import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
+import androidx.recyclerview.widget.RecyclerView;
 
 import com.joanzapata.iconify.fonts.IoniconsIcons;
 import com.randomappsinc.studentpicker.R;
 import com.randomappsinc.studentpicker.database.DataSource;
+import com.randomappsinc.studentpicker.database.NameListDataManager;
 import com.randomappsinc.studentpicker.home.MainActivity;
 import com.randomappsinc.studentpicker.models.ListInfo;
 import com.randomappsinc.studentpicker.utils.NameUtils;
 import com.randomappsinc.studentpicker.utils.UIUtils;
 
 import java.util.List;
+import java.util.Map;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
 import butterknife.Unbinder;
 
-public class GroupingMakingFragment extends Fragment {
+public class GroupingMakingFragment extends Fragment
+        implements GroupsMakingAdapter.Delegate, NameListDataManager.Listener {
 
     public static GroupingMakingFragment getInstance(String listName) {
         Bundle bundle = new Bundle();
@@ -40,11 +43,14 @@ public class GroupingMakingFragment extends Fragment {
     }
 
     @BindView(R.id.no_groups) TextView noGroups;
+    @BindView(R.id.groups_list) RecyclerView groupsList;
 
     private GroupingSettings settings;
     private GroupingSettingsDialog settingsDialog;
     private String listName;
+    private NameListDataManager nameListDataManager = NameListDataManager.get();
     private ListInfo listInfo;
+    private GroupsMakingAdapter groupsMakingListAdapter;
     private Unbinder unbinder;
 
     @Override
@@ -60,8 +66,11 @@ public class GroupingMakingFragment extends Fragment {
 
         listName = getArguments().getString(MainActivity.LIST_NAME_KEY, "");
         listInfo = new DataSource(getContext()).getListInfo(listName);
+        nameListDataManager.registerListener(this);
 
-        noGroups.setVisibility(View.VISIBLE);
+        groupsMakingListAdapter = new GroupsMakingAdapter(this);
+        groupsList.setAdapter(groupsMakingListAdapter);
+        setNoGroup();
 
         return rootView;
     }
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettingsViewHolder.java
```diff
@@ -20,10 +20,17 @@
     GroupingSettingsViewHolder(View view, GroupingSettings settings) {
         ButterKnife.bind(this, view);
         this.settings = settings;
+        numberOfNamesInList.setText(numberOfNamesInList.getContext().getResources()
```
Gear61, line 23: For some reason, getString() has priority in AOSP so it can skip the resources.
```suggestion
        numberOfNamesInList.setText(numberOfNamesInList.getContext().
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsMakingAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+                }
+                groupNames.append(NameUtils.getPrefix(i));
+                groupNames.append(listOfGroups.get(getAdapterPosition()).get(i));
+            }
+        }
+    }
+
+    void setData(List<List<String>> listOfGroups) {
+        this.listOfGroups.clear();
+        this.listOfGroups = listOfGroups;
```
Gear61: There's no point clearing the old list if we are just going to drop it and assign the reference to a new one. Let's conserve the memory allocation.
```suggestion
        this.listOfGroups.addAll(listOfGroups);
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsMakingAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+                if (i != 0) {
+                    groupNames.append("\n");
+                }
+                groupNames.append(NameUtils.getPrefix(i));
+                groupNames.append(listOfGroups.get(getAdapterPosition()).get(i));
+            }
+        }
+    }
+
+    void setData(List<List<String>> listOfGroups) {
```
Gear61: nit: Let's move this underneath the constructor, above all the RecyclerView boilerplate, since this is arguably the most important method in the adapter.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsMakingAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+import butterknife.ButterKnife;
+
+public class GroupsMakingAdapter extends RecyclerView.Adapter<GroupsMakingAdapter.GroupViewHolder> {
+
+    public interface Delegate {
+        void setNoGroup();
+    }
+
+    private List<List<String>> listOfGroups = new ArrayList<>();
+    private Delegate delegate;
```
Gear61: Delegate pattern seems heavy and unnecessary here. For now, this adapter is read only with only 1 mutation API; it just shows data. You can't click on things within the list to mutate it like you can with the name choosing adapter. Whenever the fragment calls setData(), it can just refresh the views itself afterwards.

app/src/main/res/layout/group_list_item.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.cardview.widget.CardView
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:orientation="vertical"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:layout_marginTop="2dp"
+    android:layout_marginBottom="2dp"
+    app:cardUseCompatPadding="true">
```
Gear61:
```suggestion
    app:cardUseCompatPadding="true"
    >
```

app/src/main/res/layout/group_list_item.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.cardview.widget.CardView
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:orientation="vertical"
```
Gear61: Only LinearLayouts support orientation, and I'm very sure CardView doesn't extend it.
```suggestion
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingSettings.java
```diff
@@ -2,13 +2,12 @@
 
 public class GroupingSettings {
 
+    private int listSize;
```
Gear61: nit: listSize is a little too generic for me.
```suggestion
    private int namelistSize;
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
```diff
@@ -70,27 +79,62 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
         settings = new GroupingSettings(
+                listInfo.getNumNames(),
                 getResources().getInteger(R.integer.default_number_of_names_per_group),
                 getResources().getInteger(R.integer.default_number_of_groups));
         settingsDialog = new GroupingSettingsDialog(getActivity(), settings);
     }
 
+    @Override
+    public void onNameAdded(String name, int amount, String listName) {
+        listInfo = new DataSource(getContext()).getListInfo(listName);
```
Gear61: Instead of creating a new DataSource every time, I think it's worth just having a member variable. It will save calls to getContext() as well.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
```diff
@@ -70,27 +79,62 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+        settingsDialog.refreshSetting();
+    }
+
+    @Override
+    public void onNameChanged(String oldName, String newName, int amount, String listName) {
+        listInfo = new DataSource(getContext()).getListInfo(listName);
+    }
+
+    @Override
+    public void onNameListsImported(Map<String, Integer> nameAmounts, String listName) {}
```
Gear61: We definitely need to refresh the ListInfo here. This action mutates the list.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
```diff
@@ -70,27 +79,62 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+    @Override
+    public void onNameAdded(String name, int amount, String listName) {
+        listInfo = new DataSource(getContext()).getListInfo(listName);
+        settings.setListSize(listInfo.getNumNames());
+        settingsDialog.refreshSetting();
+    }
+
+    @Override
+    public void onNameDeleted(String name, int amount, String listName) {
+        listInfo = new DataSource(getContext()).getListInfo(listName);
```
Gear61: Instead of fetching from DataSource, we can just do what NameAdapter does (check out the code) and mutate the ListInfo we have. This will save a DB read, which is pricey.

Gear61 (changes requested):
This is a great start! The cards look awesome.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsMakingAdapter.java
mohammedhemaid: can we just assign the reference instead of adding the new list? I think this will save time and memory

app/src/main/res/layout/group_list_item.xml
mohammedhemaid: Yes, the parent was a LinearLayout and I forgot to remove it after changing it to CardView.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
```diff
@@ -70,27 +79,62 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
+                listInfo.getNumNames(),
                 getResources().getInteger(R.integer.default_number_of_names_per_group),
                 getResources().getInteger(R.integer.default_number_of_groups));
         settingsDialog = new GroupingSettingsDialog(getActivity(), settings);
     }
 
+    @Override
+    public void onNameAdded(String name, int amount, String listName) {
+        listInfo = new DataSource(getContext()).getListInfo(listName);
+        settings.setListSize(listInfo.getNumNames());
```
mohammedhemaid: @Gear61 Should I put the getNumInstances() or getNumNames() the getNumNames() will not count the duplicated names and what we will do about this issue if the user puts a duplicated names in the list should we group them

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupsMakingAdapter.java
mohammedhemaid: @Gear61 What if we keep it like this instead of copying all the new values, I think this will save us time and resources.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupingMakingFragment.java
Gear61: Good question! We should use getNumInstances(). We shouldn't flatten the names when there's multiple instances I think; that doesn't feel like expected behavior.

Gear61 (approved):
Looks good!

PR #122, opened 2020-02-10
Getting rid of Name(s) deleted hack

alaaalzibda:


app/src/main/res/values/edit_name_list_strings.xml
```diff
@@ -41,7 +41,7 @@
           You currently have %d copies of this name.
      </string>
 
-    <string name="names_deleted">Name(s) deleted.</string>
+    <string name="names_deleted">%d instances of \"%s\" deleted from the list</string>
```
Gear61:
```suggestion
    <string name="names_deleted">%d instances of \"%s\" deleted from the list.</string>
```

Gear61 (changes requested):
Can you add a screenshot/video demo of both cases as well?

alaaalzibda:
Screenshot_20200213-095234
Screenshot_20200213-095248

app/src/main/res/values/edit_name_list_strings.xml
alaaalzibda: Sure!

Gear61: Can you add the period at the end? We'll be good to go after that.

Gear61 (approved):
Awesome!

PR #123, opened 2020-02-13
Automatically fill the number of groups and number of names

mohammedhemaid:
The video of the Feature:
https://photos.app.goo.gl/nKTaTJymbtFsGAET9

There is a bug I've noticed if the # names per group are empty it will change the # of groups value but it will generate groups as the old value. 
I've tested it before the commit and it worked good but I don't why the bug happened again.

The Bug is in line 57 I should set the new value of # of groups there, though it should be solved by preventing the user from setting an empty text or 0 number


app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +75,35 @@ private void applyNumberOfGroups() {
         }
         numGroups.clearFocus();
     }
+
+    @Override
+    public void beforeTextChanged(CharSequence s, int start, int count, int after) {}
```
Gear61: This is one of the very powerful parts of ButterKnife. You can use it to bind to view listening callbacks. This is how you listen to text changes without implementing 3 methods where you only use 1:
```
@OnTextChanged(value = R.id.search_input, callback = OnTextChanged.Callback.AFTER_TEXT_CHANGED)
public void afterTextChanged(Editable input) {
```
Just swap in the value with the EditText view you're binding to.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +75,35 @@ private void applyNumberOfGroups() {
...
+    @Override
+    public void afterTextChanged(Editable editable) {
+        if (isEditableAndFocused(editable, namesPerGroup)) {
+            numGroups.setText(getOffset(namesPerGroup));
+        } else if (isEditableAndFocused(editable, numGroups)) {
+            namesPerGroup.setText(getOffset(numGroups));
+        }
+    }
+
+    private String getOffset(EditText editText) {
```
Gear61: It is weird that we do things on empty input. When the input is empty or 0, I think we should just do nothing. We should only act if the enter in a number that's larger than 1.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +75,35 @@ private void applyNumberOfGroups() {
...
+        if (newNumber <= 0) {
+            newNumber = editText.getContext().getResources()
+                    .getInteger(R.integer.default_number_of_names_per_group);
+        }
+        int offset = (int) Math.ceil((double) settings.getNameListSize() / newNumber);
+        return String.valueOf(offset);
+    }
+
+    private boolean isEditableAndFocused(Editable editable, EditText editText) {
+        return editable == editText.getEditableText() && editText.isFocused();
```
Gear61: Ah, this is the struggle of binding to 2 different EditTexts with the same listener. With ButterKnife, we will be able to have 2 separate and custom named methods to listen to these callbacks. However, we will need to prevent infinite chaining. So if we change the text in number of groups, we will automatically change the text in names per group. When the callback fires for names per group, we will just early return since this EditText is not in focus.

Gear61 (changes requested):
Excited to see this feature roll out! Love the awesome and thorough test plan.

mohammedhemaid:
The video after fixed:
https://photos.app.goo.gl/mWzt5dQycqnr792S6

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +74,32 @@ private void applyNumberOfGroups() {
         }
         numGroups.clearFocus();
     }
+
+    @OnTextChanged(value = R.id.num_of_names_per_group, callback = OnTextChanged.Callback.AFTER_TEXT_CHANGED)
+    void afterNamesPerGroupChanged() {
+        String namesNumber = namesPerGroup.getText().toString().trim();
```
Gear61:
```suggestion
        String namesPerGroupInput = namesPerGroup.getText().toString().trim();
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +74,32 @@ private void applyNumberOfGroups() {
...
+        if (namesPerGroup.isFocused() && isValid(namesNumber)) {
+            numGroups.setText(getOffset(namesNumber));
+        }
+        dialog.getActionButton(DialogAction.POSITIVE).setEnabled(isValid(namesNumber));
```
Gear61, line 84: Can't we move this into the if statement?

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +74,32 @@ private void applyNumberOfGroups() {
...
+    }
+
+    @OnTextChanged(value = R.id.num_of_groups, callback = OnTextChanged.Callback.AFTER_TEXT_CHANGED)
+    void afterGroupNumChanged() {
+        String groupsNumber = numGroups.getText().toString().trim();
```
Gear61:
```suggestion
        String numberOfGroupsInput = numGroups.getText().toString().trim();
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +74,32 @@ private void applyNumberOfGroups() {
...
+        if (numGroups.isFocused() && isValid(groupsNumber)) {
+            namesPerGroup.setText(getOffset(groupsNumber));
+        }
+        dialog.getActionButton(DialogAction.POSITIVE).setEnabled(isValid(groupsNumber));
```
Gear61, line 93: Can we move this into the if statement?

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +74,32 @@ private void applyNumberOfGroups() {
...
+    }
+
+    private String getOffset(String inputtedNumber) {
```
Gear61:
```suggestion
    // Given the number of names per group or number of groups, returns the corresponding number to use all the names
    private String getMatchingNumber(String inputtedNumber) {
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -70,4 +74,32 @@ private void applyNumberOfGroups() {
...
+        int newNumber = Integer.parseInt(inputtedNumber);
+        int offset = (int) Math.ceil((double) settings.getNameListSize() / newNumber);
+        return String.valueOf(offset);
+    }
+
+    private boolean isValid(String inputText) {
```
Gear61:
```suggestion
    // Returns true if the inputted text is a positive number
    private boolean isInputValid(String inputText) {
```

app/src/main/res/values/group_setting_style.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <style name="groups_number_setting">
+        <item name="android:layout_width">wrap_content</item>
+        <item name="android:layout_height">wrap_content</item>
+        <item name="android:imeOptions">actionDone</item>
+        <item name="android:singleLine">true</item>
+        <item name="android:gravity">center</item>
+        <item name="android:inputType">number</item>
+        <item name="android:layout_marginLeft">5dp</item>
```
Gear61: Can we try to use multiples of 4 for all the spacing? So 5 -> 4, 3 -> 4, 10 -> 8 or 12.

app/src/main/res/values/group_setting_style.xml
```diff
@@ -0,0 +1,19 @@
...
+        <item name="android:paddingTop">3dp</item>
+        <item name="android:paddingBottom">3dp</item>
+        <item name="android:paddingLeft">10dp</item>
+        <item name="android:paddingRight">10dp</item>
+        <item name="android:background">@drawable/text_input_border</item>
+        <item name="android:textSize">16dp</item>
```
Gear61:
```suggestion
        <item name="android:textSize">15dp</item>
```

app/src/main/res/layout/group_making_settings.xml
```diff
@@ -60,7 +60,7 @@
         <EditText
             android:id="@+id/num_of_groups"
             android:text="@string/default_number_of_groups"
-            style="@style/two_number_setting"
+            style="@style/groups_number_setting"
```
Gear61: Why don't we apply the new style to the other EditText as well?

Gear61 (changes requested):

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
mohammedhemaid: This will make it true and it will never get false.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
mohammedhemaid: This will make it true and it will never get false.

app/src/main/res/layout/group_making_settings.xml
mohammedhemaid: I don't think the numberOfNamesPerGroup may be more than 99 names per group

app/src/main/res/values/group_setting_style.xml
```diff
@@ -0,0 +1,19 @@
...
+        <item name="android:maxLength">4</item>
```
mohammedhemaid: @Gear61 I made the length to 4 for the numOfGroups and keep it to 2 for the namesPerGroup

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
Gear61: Right, good catch!

app/src/main/res/values/group_setting_style.xml
Gear61: Ah, 2 is consistent with the choosing page. I feel like it's weird to have inconsistency; we should just cap to 2 in all places. With 4 digits, you need to deal with the EditText getting very large.

mohammedhemaid:
The video After changing the group_setting_style:
https://photos.app.goo.gl/bHZxYp9jbsq6Kefp9 
So I returned back to the two_number_setting and deleted the group setting style.

app/src/main/res/layout/group_making_settings.xml
Gear61:
```suggestion
            style="@style/two_number_setting"
```

Gear61 (approved):

PR #125, opened 2020-03-03
Added empty list and uneven groups warnings

mohammedhemaid:
The video about the feature:

https://photos.app.goo.gl/3eY2DYQVJ39mjKYEA

I know this pull will be for the variables names it's not easy to name the variables


app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingFragment.java
```diff
@@ -8,6 +8,7 @@
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.TextView;
+import android.widget.Toast;
```
mohammedhemaid: Sorry about this import I've deleted it.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingFragment.java
```diff
@@ -113,6 +114,7 @@ public void onNameListsImported(Map<String, Integer> nameAmounts, String listNam
     @OnClick(R.id.make_groups)
     void makeGroups() {
         if (listInfo.getNumNames() == 0) {
+        UIUtils.showLongToast(R.string.group_no_names_error_message,getContext());
```
Gear61: nit: Spacing is a bit off.
```suggestion
            UIUtils.showLongToast(R.string.group_no_names_error_message,getContext());
```

app/src/main/res/values/grouping_strings.xml
```diff
@@ -12,4 +12,6 @@
     <string name="num_of_names_in_group"># of names per group:</string>
     <string name="num_of_groups"># of groups:</string>
     <string name="group_number">Group # %d</string>
+    <string name="group_no_names_error_message">You have no names in the list to make groups with.</string>
+    <string name="groups_warning_message">This setting will create an uneven groups.</string>
```
Gear61: nit: Just a small wording tweak.
```suggestion
    <string name="groups_warning_message">These settings will create uneven groups.</string>
```

app/src/main/res/layout/group_making_settings.xml
```diff
@@ -63,4 +63,17 @@
...
+
+    <TextView
+        android:id="@+id/groups_warning_message"
+        android:text="@string/groups_warning_message"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginLeft="4dp"
+        android:layout_marginRight="4dp"
+        android:layout_marginBottom="4dp"
+        android:textSize="14sp"
```
Gear61: nit: Try to use 13, 15, and 17sp to stay consistent with the rest of the app
```suggestion
        android:textSize="15sp"
```

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -98,8 +100,16 @@ void afterGroupNumChanged() {
     // Given the number of names per group or number of groups, returns the corresponding number to use all the names
     private String getMatchingNumber(String inputtedNumber) {
         int newNumber = Integer.parseInt(inputtedNumber);
-        int offset = (int) Math.ceil((double) settings.getNameListSize() / newNumber);
-        return String.valueOf(offset);
+        double offset = (double) settings.getNameListSize() / newNumber;
+        checkGroupBalance(offset);
```
Gear61: In general, we want to do single responsibility design. So every module, class, and API (like this one) is only responsible for 1 thing. This makes the code clean, compact, and easy to understand. By having checkGroupBalance() fire within this method that seemingly only calculates a string given its name, this almost feels like a side effect and violates single responsibility design. I feel like we should fire this in the onAfterTextChanged() block for each of the EditTexts. It will lead to redundant calls, but this operation is cheap and code readability is more important.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -98,8 +100,16 @@ void afterGroupNumChanged() {
...
+        int matchingNumber = (int) Math.ceil(offset);
+        return String.valueOf(matchingNumber);
+    }
+
+    // check if the groups have the same names quantity
+    private void checkGroupBalance(double ratio) {
```
Gear61: We should change this method to something like maybeShowUnevenGroupsWarning(). It can extract the numbers from the views itself and then do the math. Make sure that it doesn't do anything if one of the values is empty.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -98,8 +100,16 @@ void afterGroupNumChanged() {
...
+        int warningMessageVisibility = ratio % 1 > 0 ? View.VISIBLE : View.GONE;
```
Gear61: From my understand, there's the list size and then the total number of names (# of groups * # of names per group). I think it's a legitimate use case to generate less names than the list size you have, something like generating 5 groups of 2 from a list of 60 names. I think we should only show the warning if the settings will use up all the names but not have enough.

The % 1 trick is quite nifty though, even if it's not the optimal way.

Gear61 (changes requested):
The UI looks nice, keep it up!

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
mohammedhemaid: if we do something like generating 5 groups of 2 from a list of 60 how should the autocomplete work, I think we should put a checkbox for autocomplete in the settings.

mohammedhemaid:
The Video of the new changes:
https://photos.app.goo.gl/MHK3xz5T9DDBHwLB6

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -9,16 +11,21 @@
...
 
     @BindView(R.id.number_of_names_in_list) TextView numberOfNamesInList;
     @BindView(R.id.num_of_names_per_group) EditText namesPerGroup;
     @BindView(R.id.num_of_groups) EditText numGroups;
+    @BindView(R.id.groups_warning_message) TextView warningMessage;
+    @BindView(R.id.autocomplete) CheckBox autocomplete;
 
     private GroupMakingSettings settings;
     private MaterialDialog dialog;
+    private String namesPerGroupInput;
```
Gear61: In general, it's better to stay away from having member variables that reflect information that's in the UI (which is the source of truth to the user). It's extra work maintaining these variables and easy to create bugs where what the user sees doesn't match what the business logic sees and is trying to do. If we need to get these values, we should use read them straight from the EditText.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -75,35 +82,64 @@ private void applyNumberOfGroups() {
         numGroups.clearFocus();
     }
 
+    @OnCheckedChanged(R.id.autocomplete)
+    void onAutocompleteSelected(boolean checked) {
+        if (checked) {
```
Gear61: Nesting is one of the easiest ways we can make code less readable, so let's merge this if statement and the one below it with &&, so we only have 1 layer of nesting.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -75,35 +82,64 @@ private void applyNumberOfGroups() {
...
+        }
+    }
+
     @OnTextChanged(value = R.id.num_of_names_per_group, callback = OnTextChanged.Callback.AFTER_TEXT_CHANGED)
     void afterNamesPerGroupChanged() {
-        String namesPerGroupInput = namesPerGroup.getText().toString().trim();
-        boolean isInputValid = isInputValid(namesPerGroupInput);
-        if (namesPerGroup.isFocused() && isInputValid) {
+        namesPerGroupInput = namesPerGroup.getText().toString().trim();
+        boolean isInputValid = isInputValid();
```
Gear61: Shouldn't we compute this after the autocomplete potentially happens? Like if 1 input is empty, this will evaluate to false, but then the autocomplete will fill in the other input (making it valid). But since this was calculated before the autocomplete change, the positive button gets turned off.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -75,35 +82,64 @@ private void applyNumberOfGroups() {
...
     private String getMatchingNumber(String inputtedNumber) {
         int newNumber = Integer.parseInt(inputtedNumber);
-        int offset = (int) Math.ceil((double) settings.getNameListSize() / newNumber);
-        return String.valueOf(offset);
+        double offset = (double) settings.getNameListSize() / newNumber;
+        int matchingNumber = (int) Math.ceil(offset);
+        return String.valueOf(matchingNumber);
+    }
+
+    // show warning if the groups don't have the same names quantity
```
Gear61: nit: I think we can delete this comment; the method is really well named already.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -75,35 +82,64 @@ private void applyNumberOfGroups() {
...
+        } else {
+            warningMessage.setVisibility(View.GONE);
+        }
     }
 
     // Returns true if the inputted text is a positive number
-    private boolean isInputValid(String inputText) {
-        return inputText.length() > 0 && Integer.parseInt(inputText) > 0;
+    private boolean isInputValid() {
+        return (namesPerGroupInput != null && numberOfGroupsInput != null)
```
Gear61: We can use TextUtils.isEmpty() to check for both empty AND null. It's a nice little utility method built into Android :)

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
```diff
@@ -75,35 +82,64 @@ private void applyNumberOfGroups() {
...
+        maybeShowUnevenGroupsWarning();
     }
 
     @OnTextChanged(value = R.id.num_of_groups, callback = OnTextChanged.Callback.AFTER_TEXT_CHANGED)
     void afterGroupNumChanged() {
-        String numberOfGroupsInput = numGroups.getText().toString().trim();
-        boolean isInputValid = isInputValid(numberOfGroupsInput);
-        if (numGroups.isFocused() && isInputValid) {
+        numberOfGroupsInput = numGroups.getText().toString().trim();
+        boolean isInputValid = isInputValid();
```
Gear61: Same feedback as above. The logic should be as follows:
1. Parse out the input
2. See if the autocomplete is turned on and act accordingly
3. Turn the action button on/off

Gear61 (changes requested):

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettingsViewHolder.java
mohammedhemaid: If the autocomplete change the value the isInputValid will be true since the autocomplete will fire the OnTextChanged and all the input will be valid.

mohammedhemaid:
The video of the changes:
https://photos.app.goo.gl/oB2QdfnUf2vrDx1Y8

app/build.gradle
```diff
@@ -5,7 +5,7 @@ android {
 
     def versionMajor = 3
     def versionMinor = 11
-    def versionPatch = 2
+    def versionPatch = 3
```
mohammedhemaid: Just Resolved a merge conflict

Gear61 (approved):

PR #134, opened 2020-04-21
Add import from csv option to bottom sheet

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/Cj5X5LKvkfhneTvY7

The premium version needs to restart the app right?

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
@@ -5,6 +5,7 @@ import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
 import android.view.View;
+import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 
@@ -35,6 +36,7 @@ public class HomeActivity extends StandardActivity implements
 
     @BindView(R.id.bottom_navigation) BottomNavigationView bottomNavigation;
     @BindView(R.id.bottom_sheet) View bottomSheet;
+    @BindView(R.id.title_import_from_csv) TextView importFromCsv;
 
     private HomepageFragmentController navigationController;
     private CreateListDialog createListDialog;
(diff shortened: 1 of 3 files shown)


Gear61:
@mohammedhemaid A regular user won't have access to that dialog flow. So we could update the text in onResume() since buying premium happens in an activity.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -76,6 +78,10 @@ public void onSlide(@NonNull View bottomSheet, float slideOffset) {
             showPleaseRateDialog();
         }
 
+        importFromCsv.setText(preferencesManager.isOnFreeVersion() ?
+                getString(R.string.import_from_csv_file_premium) :
```
Gear61: TextView can take in a @StringRes directly, so we can remove the getString() for more compact code.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -76,6 +78,10 @@ public void onSlide(@NonNull View bottomSheet, float slideOffset) {
             showPleaseRateDialog();
         }
 
+        importFromCsv.setText(preferencesManager.isOnFreeVersion() ?
```
Gear61: Instead of doing this in onCreate(), let's do this in onResume(). This way, when the users returns from the "Buy Premium" page having bought premium, the UI will automatically update. It will do redundant text loads, but I don't know a great way around that minus some sort of EventBus or singleton pub/sub hack. :/

Gear61 (changes requested):
A few small changes but this looks solid overall!

Gear61 (changes requested):
Code looks good, just have an issue with version control. Master should always reflect what's in production and therefore be a source of truth for everyone to work off of. Since this import from .csv feature won't land all at once, we should commit to an interstitial branch. I will create an import-from-csv branch on the core repo now. You can mirror your master to import-from-csv on your fork and we can submit a new pull request (which I'll automatically accept).

Gear61:
Actually, came up with a hack to make this faster. I can merge this in and then revert master. import-from-csv will have the new changes. I'll do that now. This will save you time.

PR #138, opened 2020-04-28
open Import Activity when file chosen

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/U4yN7XxXxL9A1ndKA

The first video I have selected is a CSV but I changed the type from .csv to .txt to see what will happen.
Also, the app will crash if I do the same for files like .png

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
@@ -16,7 +16,7 @@ import com.randomappsinc.studentpicker.common.Constants;
 import com.randomappsinc.studentpicker.common.StandardActivity;
 import com.randomappsinc.studentpicker.database.DataSource;
 import com.randomappsinc.studentpicker.editing.EditNameListActivity;
-import com.randomappsinc.studentpicker.importdata.ImportFromTextFileActivity;
+import com.randomappsinc.studentpicker.importdata.ImportFromFileActivity;
 import com.randomappsinc.studentpicker.models.ListDO;
 import com.randomappsinc.studentpicker.payments.BuyPremiumActivity;
 import com.randomappsinc.studentpicker.payments.PaymentManager;
@@ -28,13 +28,15 @@ import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
 
+import static com.randomappsinc.studentpicker.importdata.ImportFromFileActivity.IS_TEXT_FILE;
+
 public class HomeActivity extends StandardActivity implements
         BottomNavigationView.Listener, CreateListDialog.Listener, PaymentManager.Listener {
 
     private static final int NUM_APP_OPENS_FOR_RATING_ASK = 5;
 
     private static final int IMPORT_FILE_REQUEST_CODE = 1;
-    private static final int CSV_IMPORT_REQUEST_CODE = 2;
+    private static final int IMPORT_CSV_REQUEST_CODE = 2;
 
     @BindView(R.id.bottom_navigation) BottomNavigationView bottomNavigation;
     @BindView(R.id.bottom_sheet) View bottomSheet;
@@ -150,7 +152,7 @@ public class HomeActivity extends StandardActivity implements
         hideBottomSheet();
         Intent txtFileIntent = new Intent(Intent.ACTION_OPEN_DOCUMENT);
         txtFileIntent.addCategory(Intent.CATEGORY_OPENABLE);
-        txtFileIntent.setType("text/*");
+        txtFileIntent.setType("text/plain");
         startActivityForResult(txtFileIntent, IMPORT_FILE_REQUEST_CODE);
     }
 
@@ -170,7 +172,7 @@ public class HomeActivity extends StandardActivity implements
         csvIntent.addCategory(Intent.CATEGORY_OPENABLE);
         csvIntent.setType("*/*");
         csvIntent.putExtra(Intent.EXTRA_MIME_TYPES, Constants.CSV_MIME_TYPES);
-        startActivityForResult(csvIntent, CSV_IMPORT_REQUEST_CODE);
+        startActivityForResult(csvIntent, IMPORT_CSV_REQUEST_CODE);
     }
 
     private void hideBottomSheet() {
@@ -208,20 +210,27 @@ public class HomeActivity extends StandardActivity implements
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
         if (requestCode == IMPORT_FILE_REQUEST_CODE) {
-            if (resultCode == Activity.RESULT_OK && data != null && data.getData() != null) {
-                Uri uri = data.getData();
+            maybeOpenImportFileActivity(resultCode,data,true);
+        } else if (requestCode == IMPORT_CSV_REQUEST_CODE) {
+            maybeOpenImportFileActivity(resultCode,data,false);
+        }
+    }
 
-                // Persist ability to read from this file
-                int takeFlags = data.getFlags()
-                        & (Intent.FLAG_GRANT_READ_URI_PERMISSION
-                        | Intent.FLAG_GRANT_WRITE_URI_PERMISSION);
-                getContentResolver().takePersistableUriPermission(uri, takeFlags);
+    private void maybeOpenImportFileActivity(int resultCode, Intent data, boolean isTextFile) {
+        if (resultCode == Activity.RESULT_OK && data != null && data.getData() != null) {
+            Uri uri = data.getData();
 
-                String uriString = uri.toString();
-                Intent intent = new Intent(this, ImportFromTextFileActivity.class);
-                intent.putExtra(Constants.FILE_URI_KEY, uriString);
-                startActivity(intent);
-            }
+            // Persist ability to read from this file
+            int takeFlags = data.getFlags()
+                    & (Intent.FLAG_GRANT_READ_URI_PERMISSION
+                    | Intent.FLAG_GRANT_WRITE_URI_PERMISSION);
+            getContentResolver().takePersistableUriPermission(uri, takeFlags);
+
+            String uriString = uri.toString();
+            Intent intent = new Intent(this, ImportFromFileActivity.class);
+            intent.putExtra(Constants.FILE_URI_KEY, uriString);
+            intent.putExtra(IS_TEXT_FILE, isTextFile);
+            startActivity(intent);
         }
     }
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -28,13 +28,15 @@
...
 import butterknife.OnClick;
 
+import static com.randomappsinc.studentpicker.importdata.ImportFromFileActivity.IS_TEXT_FILE;
+
 public class HomeActivity extends StandardActivity implements
         BottomNavigationView.Listener, CreateListDialog.Listener, PaymentManager.Listener {
 
     private static final int NUM_APP_OPENS_FOR_RATING_ASK = 5;
 
     private static final int IMPORT_FILE_REQUEST_CODE = 1;
```
Gear61: Let's make the old code more readable and able to support our change.
```suggestion
    private static final int IMPORT_TXT_REQUEST_CODE = 1;
```

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -208,20 +210,27 @@ protected void onResume() {
...
-                intent.putExtra(Constants.FILE_URI_KEY, uriString);
-                startActivity(intent);
-            }
+            maybeOpenImportFileActivity(resultCode,data,true);
+        } else if (requestCode == IMPORT_CSV_REQUEST_CODE) {
+            maybeOpenImportFileActivity(resultCode,data,false);
+        }
+    }
+
+    private void maybeOpenImportFileActivity(int resultCode, Intent data, boolean isTextFile) {
```
Gear61: Even though there are only 2 cases, equating (not .txt) with (is .csv) is hard to read and won't scale if we support more import types. Let's create an [IntDef](https://developer.android.com/reference/android/support/annotation/IntDef) for the 2 file types. We can name it FileImportType and put it either in common or importdata. We will need to change the method signature of this method alongside the data passed between activities via intent.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -208,20 +210,27 @@ protected void onResume() {
...
-                        & (Intent.FLAG_GRANT_READ_URI_PERMISSION
-                        | Intent.FLAG_GRANT_WRITE_URI_PERMISSION);
-                getContentResolver().takePersistableUriPermission(uri, takeFlags);
-
-                String uriString = uri.toString();
-                Intent intent = new Intent(this, ImportFromTextFileActivity.class);
-                intent.putExtra(Constants.FILE_URI_KEY, uriString);
-                startActivity(intent);
-            }
+            maybeOpenImportFileActivity(resultCode,data,true);
```
Gear61: Small spacing thing.
```suggestion
            maybeOpenImportFileActivity(resultCode, data, true);
```

app/src/main/java/com/randomappsinc/studentpicker/importdata/ImportFromFileActivity.java
```diff
@@ -54,7 +58,7 @@ public void onConfigurationChanged(@NonNull Configuration newConfig) {
         bannerAdManager.onOrientationChanged();
     }
 
-    private void extractNameListInfo() {
+    private void extractNameListInfoFromTextFile() {
```
Gear61: I was lazy and inlined all the file parsing logic within the activity. I think as we build this out, we should create a FileParsingManager class. We pass it the file to parse and it parses on the background thread. When it's done, it lets the activity know what it parsed via a callback.

Gear61 (changes requested):

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/nwMNySqauEZGdrPi7

What we should do next?
If we have a big file and someone tries to exit the ImportFileAtivity before the import is finished we won't be able to save all the names.
so we have two options:
1- create a warning dialog so if someone tries to exit the ImportFileAtivity while importing, the list will not be saved and we need to delete the list from the database, because some names will be already added
2- create a service or job scheduler to continue importing the name in the background.

BUG
NameChoosenActiviy will have some lag if we navigate back.
The video of the bug:
https://photos.app.goo.gl/VDWLhq6eFo8GQcGx7

So should we make a pagination?

app/src/main/java/com/randomappsinc/studentpicker/importdata/FileParsingManager.java
```diff
@@ -0,0 +1,172 @@
...
+                        reader.close();
+                    }
+                } catch (IOException ignored) {}
+            }
+
+            listener.onFileParsingSuccess(listNameText, namesText.toString());
+        });
+    }
+
+    void saveNameList(Context context, String newListName, String[] allNames) {
```
Gear61: With this method, FileParsingManager as a name doesn't capture the scope of this class anymore. I think we should rename this to FileImportManager.

app/src/main/java/com/randomappsinc/studentpicker/importdata/ImportFromFileActivity.java
```diff
@@ -58,55 +62,26 @@ public void onConfigurationChanged(@NonNull Configuration newConfig) {
...
-                    if (inputStream != null) {
-                        inputStream.close();
-                    }
-                    if (reader != null) {
-                        reader.close();
-                    }
-                } catch (IOException ignored) {}
-            }
-        });
+    private void extractNamesList(int fileType) {
```
Gear61: We should make the activity even dumber. Just pass in the file type to the FileImportManager and have it figure out which parsing logic to use.

app/src/main/java/com/randomappsinc/studentpicker/importdata/ImportFromFileActivity.java
```diff
@@ -58,55 +62,26 @@ public void onConfigurationChanged(@NonNull Configuration newConfig) {
...
+                fileParsingManager.extractNameListFromText(this, fileUri);
+                break;
+            case FileImportType.CSV:
+                fileParsingManager.extractNameListFromCsv(this, fileUri);
+                break;
+        }
+    }
+
+    @Override
+    public void onFileParsingFailure(int fileParsingError) {
```
Gear61: Also, I think we should UI thread this since Toast is a UI change and this can be called on the background thread right?
```suggestion
    public void onFileParsingFailure(@StringRes int fileParsingError) {
```

app/src/main/java/com/randomappsinc/studentpicker/importdata/ImportFromFileActivity.java
```diff
@@ -58,55 +62,26 @@ public void onConfigurationChanged(@NonNull Configuration newConfig) {
...
+        UIUtils.showLongToast(fileParsingError, this);
+    }
+
+    @Override
+    public void onFileParsingSuccess(String listNameText, String namesText) {
+        loadUI(listNameText, namesText);
```
Gear61: We should just inline loadUI here. There isn't a ton of value in having a relatively small method that's only called once.

app/src/main/java/com/randomappsinc/studentpicker/importdata/ImportFromFileActivity.java
```diff
@@ -119,23 +94,23 @@ private void loadUI(String listNameText, String namesListText) {
...
-
-            String[] allNames = names.getText().toString().split("\\r?\\n");
-            for (String name : allNames) {
-                String cleanName = name.trim();
-                if (!cleanName.isEmpty()) {
-                    dataSource.addNames(cleanName, 1, newList.getId());
-                }
-            }
+            loading.setVisibility(View.VISIBLE);
+            view.setEnabled(false);
```
Gear61: Disabling the view leads to some weird cases. For example, what if the saving fails; you need to re-enable the button. I also think it's weird just having a progress bar go over the UI. What if we use a blocking MaterialDialog instead? You can do indeterminate progress dialog with it and make it non-cancelable. This will handle point #1 that you brought up. I feel like if the user is going to push the app with 1,000 names, they should be ready to wait a few seconds.

app/src/main/java/com/randomappsinc/studentpicker/importdata/FileParsingManager.java
```diff
@@ -0,0 +1,172 @@
...
+                    if (inputStream != null) {
+                        inputStream.close();
+                    }
+                    if (reader != null) {
+                        reader.close();
+                    }
+                } catch (IOException ignored) {}
+            }
+
+            listener.onFileParsingSuccess(listNameText, namesText.toString());
```
Gear61, line 153: The original text parsing code I wrote was pretty hacky. I don't like combining the model (the data) and the formatting together. I think it's also out of scope for this component to know how the UI works (how does it know that the consumer wants to data split up by new line?). I think we should just return a list of strings. The activity can take in the list of strings, store it as a member variable, and create a temp formatted string for it to load into the UI.

Gear61 (changes requested):
This looks way better than before. Thank you for going above and beyond by creating the manager class. This makes the overall code a lot cleaner.

Gear61:
@mohammedhemaid Good catch on the lag for name choosing activity when we navigate back. It's due to us persisting the list state to database when we exit. We should move that into the background thread. I feel like that's out of scope for the import from .csv feature though, so let's wrap up this feature and tackle it later.

app/src/main/java/com/randomappsinc/studentpicker/importdata/ImportFromFileActivity.java
mohammedhemaid: yes, but what if the user click save many times? also progress dialog is deprecated but it's the better solution right now.

Gear61: There's a limit to how much we should protect the UI. I feel like the user double-tapping the button before the dialog shows up is incredibly difficult. If the user wants to break the app, let them break the app.

app/src/main/java/com/randomappsinc/studentpicker/importdata/FileImportManager.java
```diff
@@ -0,0 +1,189 @@
...
+        this.fileUri = fileUri;
+
+        HandlerThread handlerThread = new HandlerThread("File parsing thread");
+        handlerThread.start();
+        backgroundHandler = new Handler(handlerThread.getLooper());
+
+        importFile(fileType);
+    }
+
+    private void importFile(int fileType) {
```
Gear61, line 52:
```suggestion
    private void importFile(@FileImportType int fileType) {
```

Gear61 (approved):
Have some feedback but just going to merge this in to unblock.

PR #139, opened 2020-05-07
Initial theme mode

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/UXhRZQybrh5bKHVz5

For the fragments recreation, we can use the old instance but that requires more work:
1- override onSavedInstanceState and save the currentViewId for the selected fragment.
2- on recreation  in OnCreate we will assign the homepageFragment and settingsFragment in the 
 HomepageFragmentController with the old instance from the FragmentManager.
3- In the BottomNavigationView we will need to select the selected fragment.

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
@@ -29,6 +29,8 @@ import butterknife.OnClick;
 public class HomeActivity extends StandardActivity implements
         BottomNavigationView.Listener, CreateListDialog.Listener, PaymentManager.Listener {
 
+    private static final String PREVIOUS_SELECTED_ID = "previousSelectedId";
+
     private static final int NUM_APP_OPENS_FOR_RATING_ASK = 5;
 
     private static final int IMPORT_FILE_REQUEST_CODE = 1;
@@ -49,7 +51,14 @@ public class HomeActivity extends StandardActivity implements
         ButterKnife.bind(this);
 
         navigationController = new HomepageFragmentController(getSupportFragmentManager(), R.id.container);
-        navigationController.loadHomeInitially();
+        if (savedInstanceState == null) {
+            navigationController.loadHomeInitially();
+        } else {
+            navigationController.restoreFragments();
+            int previousSelectedId = savedInstanceState.getInt(PREVIOUS_SELECTED_ID, R.id.home);
+            navigationController.onNavItemSelected(previousSelectedId);
+            bottomNavigation.setCurrentlySelected(previousSelectedId);
+        }
 
         bottomSheetBehavior = BottomSheetBehavior.from(bottomSheet);
         bottomSheetBehavior.setState(BottomSheetBehavior.STATE_HIDDEN);
@@ -80,6 +89,12 @@ public class HomeActivity extends StandardActivity implements
         dataSource = new DataSource(this);
         paymentManager = new PaymentManager(this, this);
         paymentManager.setUpAndCheckForPremium();
+    }
+
+    @Override
+    public void onSaveInstanceState(@NonNull Bundle outState) {
+        super.onSaveInstanceState(outState);
+        outState.putInt(PREVIOUS_SELECTED_ID, navigationController.getCurrentViewId());
     }
 
     @Override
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/randomappsinc/studentpicker/settings/SettingsFragment.java
```diff
@@ -94,6 +96,15 @@ public void onItemClick(int position) {
                 preferencesManager.setShakeEnabled(!currentState);
                 return;
             case 2:
+                int darkModePosition = preferencesManager.isOnFreeVersion() ? 2 : 1;
+                View secondCell = settingsOptions.getChildAt(darkModePosition);
+                Switch DarkModeToggle = secondCell.findViewById(R.id.shake_toggle);
```
Gear61:
```suggestion
                Switch darkModeToggle = secondCell.findViewById(R.id.shake_toggle);
```

app/src/main/res/values/settings_strings.xml
```diff
@@ -3,13 +3,15 @@
     <string name="buy_premium">Buy premium</string>
     <string name="buy_button_text">Buy ($0.99)</string>
     <string name="shake_enabled">Shake to choose</string>
+    <string name="dark_Mode">Dark Mode</string>
```
Gear61:
```suggestion
    <string name="dark_mode">Dark mode</string>
```

Gear61 (changes requested):
This looks great! The code makes a lot of sense to me, and it's quite clean overall.

Anyways, after watching the video, it's a pretty bad experience when you toggle the dark mode switch and you get taken back to the homepage. Does onSavedInstanceState() get called when the app restarts? We could theoretically call getActivity() and call some sort of setter on it so that when it restarts (savedInstanceState Bundle is non-null in onCreate()), it knows to tab to the settings fragment. Is this feasible? Let's try to make this user experience work, even if we have to write some messy code. We can always wrap the code to make it not as messy, haha.

mohammedhemaid:
The video of the update:
https://photos.app.goo.gl/BjGTYgdb5Vqq2pwf9

mohammedhemaid:
@Gear61 
I was worried about the messy code hahaha,  now I used the old instance and get the id of the latest selected view, it now working but the code still needs some cleaning.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -49,7 +51,14 @@ protected void onCreate(Bundle savedInstanceState) {
         ButterKnife.bind(this);
 
         navigationController = new HomepageFragmentController(getSupportFragmentManager(), R.id.container);
-        navigationController.loadHomeInitially();
+        if (savedInstanceState == null) {
```
Gear61, line 54: We should annotate savedInstanceState as @Nullable.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -29,6 +29,8 @@
 public class HomeActivity extends StandardActivity implements
         BottomNavigationView.Listener, CreateListDialog.Listener, PaymentManager.Listener {
 
+    private static final String PREVIOUS_SELECTED_ID = "previousSelectedId";
```
Gear61, line 32: This is more descriptive IMHO.
```suggestion
    private static final String PREVIOUSLY_SELECTED_PAGE_ID = "previouslySelectedPageId";
```

app/src/main/java/com/randomappsinc/studentpicker/home/HomepageFragmentController.java
```diff
@@ -53,10 +52,15 @@ void onNavItemSelected(@IdRes int viewId) {
...
     /** Called by the app upon start up to load the homepage */
     void loadHomeInitially() {
+        this.homepageFragment = HomepageFragment.getInstance();
         currentViewId = R.id.home;
         addFragment(homepageFragment);
     }
 
+     int getCurrentViewId() {
+        return currentViewId;
+    }
```
Gear61, line 62:
```suggestion
    int getCurrentViewId() {
        return currentViewId;
    }
```

Gear61 (approved):
Good first step! I just have some small comments you can dress in the next pull request.

PR #159, opened 2020-05-24
Update schema to add grouping settings

mohammedhemaid:


mohammedhemaid (commented):
To be honest I don't have a lot of experience with SQLite database so In line 309 how the LEGACY_TABLE_NAME is created?

Gear61 (approved):
Awesome!

PR #161, opened 2020-05-26
Persist the group making settings

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/MQe9zmBK2uuBoFyY6

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/database/MySQLiteHelper.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/database/MySQLiteHelper.java
@@ -273,7 +273,7 @@ public class MySQLiteHelper extends SQLiteOpenHelper {
             safelyAddColumnToTable(
                     database,
                     DatabaseColumns.NUMBER_OF_GROUPS,
-                    "INTEGER DEFAULT 0",
+                    "INTEGER DEFAULT 1",
                     DatabaseTables.NAMES);
         }
     }
--- a/app/src/main/java/com/randomappsinc/studentpicker/database/MySQLiteHelper.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/database/MySQLiteHelper.java
@@ -273,7 +273,7 @@ public class MySQLiteHelper extends SQLiteOpenHelper {
             safelyAddColumnToTable(
                     database,
                     DatabaseColumns.NUMBER_OF_GROUPS,
-                    "INTEGER DEFAULT 1",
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/randomappsinc/studentpicker/database/MySQLiteHelper.java
```diff
@@ -273,7 +273,7 @@ public void onUpgrade(SQLiteDatabase database, int oldVersion, int newVersion) {
             safelyAddColumnToTable(
                     database,
                     DatabaseColumns.NUMBER_OF_GROUPS,
-                    "INTEGER DEFAULT 1",
+                    "INTEGER DEFAULT 0",
```
Gear61: I feel like this should be 1? The code is resilient and doesn't crash if the settings are bogus I think.

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettings.java
```diff
@@ -1,37 +1,39 @@
...
         return numOfGroups;
     }
 
-    void setNumOfGroups(int numOfGroups) {
-        this.numOfGroups = numOfGroups;
+    public void setNumOfGroups(int numOfGroups) {
+        if (numOfGroups != 0) {
+            this.numOfGroups = numOfGroups;
+        } else {
+            int numGroupsToFill = nameListSize / DEFAULT_NAMES_PER_GROUP;
```
Gear61: I feel like we're overthinking the setter. If the user is purposely setting the value to 0, I think you should just let them do it. This code makes this API less readable and introduces complication since the 0 mutation should be reflected in the UI as well. you should just let my old code die, lol.

Gear61 (changes requested):
Looks great, just a few comments!

app/src/main/java/com/randomappsinc/studentpicker/grouping/GroupMakingSettings.java
mohammedhemaid: @Gear61 
if I deleted this code the default value will always be 1 and we will not be able to set the initial value of groups number as names list size

Gear61: Wait, is this in the DataSource layer? We can expose a constructor with 3 parameters instead right?

Gear61 (approved):
Neat.

PR #163, opened 2020-05-31
Show photos in choosing dialog

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/82obHNWVRZG2ES3s6

I am thinking to add a line divider between the names when I showing them in list


app/build.gradle
```diff
@@ -1,11 +1,14 @@
...
+apply plugin: 'kotlin-android-extensions'
+apply plugin: 'kotlin-kapt'
 
 android {
     compileSdkVersion 29
 
     def versionMajor = 4
     def versionMinor = 3
-    def versionPatch = 0
+    def versionPatch = 1
```
Gear61: Sorry, we're on 4.3.2 now :/

I needed to fix the group making settings schema migration.

app/src/main/java/com/randomappsinc/studentpicker/choosing/ChoicesDisplayViewHolder.kt
```diff
@@ -0,0 +1,30 @@
...
+        view: View,
+        private val listInfo: ListInfo
+) {
+
+    @BindView(R.id.names_list) lateinit var namesList: RecyclerView
+
+    var chooseNameDialogAdapter: ChooseNameDialogAdapter
+
+    init {
+        ButterKnife.bind(this, view)
```
Gear61: Can we not use Android View Binding instead of ButterKnife? Not a huge deal, just wondering.

app/src/main/java/com/randomappsinc/studentpicker/choosing/ChooseNameDialogAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        @BindView(R.id.person_number) lateinit var nameNumber: TextView
+        @BindView(R.id.person_image) lateinit var personImageView: ImageView
+        @BindView(R.id.person_name) lateinit var name: TextView
+
+        init {
+            ButterKnife.bind(this, view)
+        }
+
+        @SuppressLint("SetTextI18n")
```
Gear61: Interesting, does this lint not exist in Java? Can we remove this?

app/src/main/java/com/randomappsinc/studentpicker/choosing/ChooseNameDialogAdapter.kt
```diff
@@ -0,0 +1,80 @@
...
+        fun loadName(position: Int) {
+            val nameDO: NameDO = currentState.getNameDO(chosenNames[position])
+            nameNumber.maybeVisible(showAsList)
```
Gear61: Why isn't L70 the same as this one? If this the same (false) -> View.GONE, (true) -> View.VISIBLE logic?

app/src/main/java/com/randomappsinc/studentpicker/extentions/ViewExt.kt
```diff
@@ -0,0 +1,7 @@
+package com.randomappsinc.studentpicker.extentions
+
+import android.view.View
+
+fun View.maybeVisible(show: Boolean) {
```
Gear61: Ah, that's how maybeVisible works. I don't think this extension is worth it, and extensions on core classes can make them less readable over time. It's also not clear from the API name what it does - What if we want the not visible state to be View.INVISIBLE instead?

app/src/main/res/layout/name_choices_dialog.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout
```
Gear61: I don't think we need this LinearLayout. Let's try to remove it to improve performance.

app/src/main/java/com/randomappsinc/studentpicker/choosing/ChoicesDisplayDialog.java
```diff
@@ -22,13 +23,15 @@
...
+    private ChoicesDisplayViewHolder choicesDisplayViewHolder;
 
-    ChoicesDisplayDialog(Listener listener, Context context, int listId, ChoosingSettings settings) {
+    ChoicesDisplayDialog(Listener listener, Context context, int listId, ListInfo listInfo, ChoosingSettings settings) {
         this.listId = listId;
         this.settings = settings;
         dialog = new MaterialDialog.Builder(context)
                 // Placeholder because otherwise, the view doesn't exist
                 .title(R.string.name_chosen)
+                .customView(R.layout.name_choices_dialog, false)
```
Gear61: Instead of custom view, I think we can just use MaterialDialog's .adapter() API. The dialog will just provide the RecyclerView for us :)

Gear61 (changes requested):

app/src/main/java/com/randomappsinc/studentpicker/choosing/ChooseNameDialogAdapter.kt
mohammedhemaid: hmmmm this one because I am concatenating the two string and they saying that I should use the string resources instead

app/src/main/java/com/randomappsinc/studentpicker/choosing/ChooseNameDialogAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+            notifyDataSetChanged()
+        }
+
+    var showAsList = false
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
+        return ViewHolder(
+                ChooseNameDialogCellBinding.inflate(
+                        LayoutInflater.from(parent.context), parent, false))
+
```
Gear61, line 34: Delete L34?

Gear61 (approved):
Congratulations on merging the first Kotlin into Random Name Picker :)

PR #168, opened 2020-06-14
Show photos on presentation page

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/8Q5bR4xXiUooNEQF9

I created a presentation manager that has the animations with the player 

-For the text size and color dialogs, so should I remove them or we need to use them in some way?

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
@@ -1,25 +1,15 @@
 package com.randomappsinc.studentpicker.presentation;
 
-import android.animation.Animator;
-import android.animation.AnimatorSet;
-import android.animation.ObjectAnimator;
-import android.content.Context;
-import android.content.res.AssetFileDescriptor;
-import android.media.AudioManager;
-import android.media.MediaPlayer;
+import android.graphics.drawable.Drawable;
 import android.os.Bundle;
-import android.os.Handler;
-import android.util.TypedValue;
-import android.view.Gravity;
 import android.view.Menu;
 import android.view.MenuItem;
-import android.view.View;
-import android.view.animation.AccelerateInterpolator;
-import android.view.animation.DecelerateInterpolator;
 import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.recyclerview.widget.DividerItemDecoration;
+import androidx.recyclerview.widget.RecyclerView;
 
 import com.afollestad.materialdialogs.DialogAction;
 import com.afollestad.materialdialogs.MaterialDialog;
@@ -27,9 +17,11 @@ import com.afollestad.materialdialogs.color.ColorChooserDialog;
 import com.joanzapata.iconify.IconDrawable;
 import com.joanzapata.iconify.fonts.IoniconsIcons;
 import com.randomappsinc.studentpicker.R;
+import com.randomappsinc.studentpicker.common.ChooseNameAdapter;
 import com.randomappsinc.studentpicker.choosing.ChoosingSettings;
 import com.randomappsinc.studentpicker.database.DataSource;
 import com.randomappsinc.studentpicker.models.ListInfo;
+import com.randomappsinc.studentpicker.models.NameDO;
 import com.randomappsinc.studentpicker.speech.TextToSpeechManager;
 import com.randomappsinc.studentpicker.utils.NameUtils;
 import com.randomappsinc.studentpicker.utils.PreferencesManager;
@@ -38,25 +30,27 @@ import com.randomappsinc.studentpicker.utils.UIUtils;
 import java.util.List;
 
 import butterknife.BindColor;
+import butterknife.BindDrawable;
 import butterknife.BindInt;
 import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
 
 public class PresentationActivity extends AppCompatActivity
-        implements ColorChooserDialog.ColorCallback, TextToSpeechManager.Listener {
+        implements ColorChooserDialog.ColorCallback, TextToSpeechManager.Listener,
+        PresentationManager.Listener {
 
     public static final String LIST_ID_KEY = "listId";
     public static final String DRUMROLL_FILE_NAME = "drumroll.mp3";
 
     @BindView(R.id.header) TextView header;
-    @BindView(R.id.names) TextView names;
+    @BindView(R.id.names_list) RecyclerView namesList;
     @BindColor(R.color.text_normal) int textNormalColor;
     @BindInt(R.integer.default_anim_length) int defaultAnimationLengthMs;
+    @BindDrawable(R.drawable.line_divider) Drawable lineDivider;
 
     private PreferencesManager preferencesManager;
     private DataSource dataSource;
-    private MediaPlayer player;
     private int listId;
     private ListInfo listState;
     private ChoosingSettings settings;
@@ -64,8 +58,8 @@ public class PresentationActivity extends AppCompatActivity
     private MaterialDialog setTextSizeDialog;
     private SetTextSizeViewHolder setTextViewHolder;
     private TextToSpeechManager textToSpeechManager;
-    private Handler handler;
-    private Runnable animateNamesTask = this::animateNames;
+    private PresentationManager presentationManager;
+    private ChooseNameAdapter chooseNameAdapter;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
```diff
@@ -85,19 +79,17 @@ protected void onCreate(Bundle savedInstanceState) {
         settings = dataSource.getChoosingSettings(listId);
         textToSpeechManager = new TextToSpeechManager(this, this);
 
+        presentationManager = new PresentationManager(this, namesList);
+        DividerItemDecoration itemDecorator =
+                new DividerItemDecoration(this, DividerItemDecoration.VERTICAL);
```
Gear61: I am unsure how to feel about the divider. I know it's in the dialog, but on this page, it just seems... off.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
```diff
@@ -106,7 +98,6 @@ protected void onCreate(Bundle savedInstanceState) {
                 .onPositive((@NonNull MaterialDialog dialog, @NonNull DialogAction which) -> {
                     int newTextSize = setTextViewHolder.textSizeSlider.getProgress() + 1;
                     preferencesManager.setPresentationTextSize(newTextSize);
-                    names.setTextSize(TypedValue.COMPLEX_UNIT_SP, newTextSize * 8);
```
Gear61, line 109: Users get mad when you disable random features. I think we should make the text size at the adapter layer dynamic and call notifyDataSetChanged() when this happens.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationManager.kt
```diff
@@ -0,0 +1,100 @@
...
+    private val animateNamesTask = ::animateNames
+    private var defaultAnimationLengthMs: Long = 0
+
+    init {
+        handler = Handler()
+        player = MediaPlayer()
+        defaultAnimationLengthMs = view.context.resources.getInteger(R.integer.default_anim_length).toLong()
+    }
+
+    fun playSoundOrAnimations() {
```
Gear61: playSoundOrAnimations() reveals too much about the implementation, which isn't necessary. Also, if we change the nature of the presentation, we will have to rename this method. To keep it "magic"-sounding, we should just follow the name of the class. I think presentChosenName() is a fitting method name for a PresentationManager class.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationManager.kt
```diff
@@ -0,0 +1,100 @@
...
+            player!!.setDataSource(
+                    fileDescriptor.fileDescriptor,
+                    fileDescriptor.startOffset,
+                    fileDescriptor.length)
+            player!!.prepare()
+            player!!.start()
+        } catch (ex: Exception) {
+            UIUtils.showLongToast(R.string.drumroll_error, view.context)
+        }
+        playNamesAnimation(2600)
```
Gear61: nit: Let's move 2600 to some sort of constant like ANIMATION_DELAY_MS. Maybe have a comment saying that 2.6s is the amount of time you need to wait for the drumroll crash to happen (we want to sync the animating in of the names with that crash).

Gear61:
The animation looks kind of... weird as well. It looks like it pops out too much and goes really up. Maybe we should tone down the scaling to make it more subtle? Like a 10% or 25% scale.

Gear61 (changes requested):

mohammedhemaid:
The video of the changes:
https://photos.app.goo.gl/EF7hHfSPbqtzJ97P6

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
```diff
@@ -251,7 +176,7 @@ public void onPause() {
     public void finish() {
         super.finish();
         overridePendingTransition(0, R.anim.slide_out_from_top);
-        player.stop();
+        presentationManager.stopPlayer();
```
Gear61: nit: stopPlayer() made more sense when we were inlining the MediaPlayer directly. I think stopPresentation() is a better API name given the name.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationAdapter.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    inner class ViewHolder(itemBinding: ChooseNameDialogCellBinding) : RecyclerView.ViewHolder(itemBinding.root) {
+
+        private val nameNumber: TextView = itemBinding.personNumber
+        private val personImageView: ImageView = itemBinding.personImage
+        private val name: TextView = itemBinding.personName
+
+        fun loadName(position: Int) {
+            val nameDO: NameDO = chosenNames[position]
+            if (showAsList) {
```
Gear61, line 66: The vast majority of users won't have showAsList setting (it's a pretty niche setting). This adds more code complexity, but let's change the gravity of the container based on this setting to adhere to the old implementation. So if this setting is on, we keep the regular align start since we need to align the numbers. If it's off, set the gravity of the content inside to be centered horizontally. It looks really weird to only select 1 thing and have it off to the side.

Gear61 (changes requested):
The v2 looks a lot better. On a side note, all this nuance shows how allowing for user customization is a slippery slope that leads to hard to maintain code... I probably accepted too many of these requests.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationAdapter.kt
mohammedhemaid: yes, it looks very weird thanks for the feedback

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationManager.kt
```diff
@@ -0,0 +1,98 @@
...
+import android.media.AudioManager
+import android.media.MediaPlayer
+import android.os.Handler
+import android.view.View
+import android.view.animation.AccelerateInterpolator
+import android.view.animation.DecelerateInterpolator
+import com.randomappsinc.studentpicker.R
+import com.randomappsinc.studentpicker.utils.UIUtils
+
+/** The amount of time you need to wait for the drumroll crash to happen*/
```
Gear61: nit: Is it not possible to have this within the class?
```suggestion
/** The amount of time you need to wait for the drumroll crash to happen */
```

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationManager.kt
```diff
@@ -0,0 +1,98 @@
...
+        private val view: View) {
+
+    interface Listener {
+        fun speakNames()
+    }
+
+    private var player: MediaPlayer = MediaPlayer()
+    private var handler: Handler = Handler()
+    private val animateNamesTask = ::animateNames
+    private var defaultAnimationLengthMs: Long = view.context.resources.getInteger(
```
Gear61, line 33: Isn't it clearly an integer? We don't need values greater than 2 billion.
```suggestion
    private var defaultAnimationLengthMs: Integer = view.context.resources.getInteger(
```

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationManager.kt
```diff
@@ -0,0 +1,98 @@
...
+                .ofFloat(view, View.SCALE_Y, 1.0f, 1.25f, 1.0f)
+                .setDuration(defaultAnimationLengthMs)
+        scaleSet.interpolator = DecelerateInterpolator()
+        scaleSet.playTogether(scaleX, scaleY)
+        val fullSet = AnimatorSet()
+        fullSet.playSequentially(fadeIn, scaleSet)
+        fullSet.addListener(object : Animator.AnimatorListener {
+            override fun onAnimationStart(animator: Animator) {}
+            override fun onAnimationEnd(animator: Animator) {
+                listener.speakNames()
```
Gear61: This doesn't seem right - Automatically speaking the names is a toggle-able setting right? Don't we need to read from some boolean?

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationActivity.java
```diff
@@ -123,85 +111,22 @@ public void choose() {
...
-                    fileDescriptor.getFileDescriptor(),
-                    fileDescriptor.getStartOffset(),
-                    fileDescriptor.getLength());
-            player.prepare();
-            player.start();
-        } catch (Exception ex) {
-            UIUtils.showLongToast(R.string.drumroll_error, this);
+    @Override
+    public void speakNames() {
+        if (settings.getAutomaticTts()) {
```
Gear61: Ahhh, the speaking check is here. Calling speakNames() to have to sometimes not speak names is confusing and wastes a method call. I think we should just pass the boolean into presentation manager.

Gear61 (changes requested):
Just some small things - This looks great overall!

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationManager.kt
mohammedhemaid: Hmmm, we can put it inside a companion object but this is much less clean because under the hood when the bytecode is generated, a useless object is created.

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationManager.kt
mohammedhemaid: The ObjectAnimator duration is taking long, Kotlin is weird it doesn't have an implicit casting like java I think that because it doesn't have a primitive data type.

mohammedhemaid:
The video of the changes:
https://photos.app.goo.gl/csuMJrxGXMXMp9Ew6

I pushed the first commit without testing then realized that, so suddenly the app crashed while I was testing it so I fixed it.
Also changed the presentation mic icon to speaker icon, IMHO that looks good and better for the user experience

app/src/main/java/com/randomappsinc/studentpicker/presentation/PresentationAdapter.kt
```diff
@@ -0,0 +1,92 @@
...
+    inner class ViewHolder(itemBinding: ChooseNameDialogCellBinding) : RecyclerView.ViewHolder(itemBinding.root) {
+
+        private val parent: LinearLayout = itemBinding.root
+        private val nameNumber: TextView = itemBinding.personNumber
+        private val personImageView: ImageView = itemBinding.personImage
+        private val name: TextView = itemBinding.personName
+
+        fun loadName(position: Int) {
+            val nameDO: NameDO = chosenNames[position]
+            if (showAsList) {
```
Gear61, line 66: Show as list can't change while the user is on the screen, so technically you can move this into onCreateViewHolder where the view is inflated so you don't need to do it again on this view being recycled and rebound.

Gear61 (approved):
This has been in review for a long time so merging this in. Great job!

PR #170, opened 2020-07-26
Fix speech to text doesn't automatically open after accepting permission

mohammedhemaid:
The video of the fix:

https://user-images.githubusercontent.com/1650117/182970997-c5e6b3e0-cac2-4e01-a618-5f99b124ce0c.mp4

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/home/HomepageFragment.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/home/HomepageFragment.java
@@ -11,7 +11,6 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.EditText;
 import android.widget.TextView;
-import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.fragment.app.Fragment;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/randomappsinc/studentpicker/home/HomepageFragment.java
```diff
@@ -136,6 +138,18 @@ void searchWithVoice() {
...
     }
 
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+        if (requestCode == RECORD_AUDIO_PERMISSION_CODE) {
+            if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
+                speechToTextManager.startSpeechToTextFlow();
+            } else {
+                Toast.makeText(getContext(), R.string.speech_microphone_permission_denied, Toast.LENGTH_LONG).show();
```
Gear61: nit: We should use UIUtils.showLongToast(). If it doesn't exist, we should make one.

app/src/main/java/com/randomappsinc/studentpicker/home/HomepageFragment.java
```diff
@@ -136,6 +138,18 @@ void searchWithVoice() {
         }
     }
 
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+        if (requestCode == RECORD_AUDIO_PERMISSION_CODE) {
```
Gear61, line 142: Consistency across an app is vital. We should make sure that if a certain thing happens in 1 part of the app (like record audio permission being denied), the experience is the same, otherwise it will be confusing to the user. Let's make sure the speech to text flow in the edit name list activity has the same code.

app/src/main/res/values/home_strings.xml
```diff
@@ -30,4 +30,6 @@
         We detected that you purchased Random Name Picker Premium on another device.
         You can restart the app to immediately see all the upgrades.
     </string>
+
+    <string name="speech_microphone_permission_denied">Speech search requires the use of the microphone.</string>
```
Gear61: This is a more generic string that we can use in other places.
```suggestion
    <string name="speech_to_text_permission_denied">To convert your speech to text, we need access to your microphone.</string>
```

Gear61 (changes requested):

Gear61 (approved):
Looks good!

PR #172, opened 2020-08-31
Fix name photo disappear after deleted

mohammedhemaid:
It was a small fix. 

The only bug right now if we consider it as a bug is we are having two copies of the image that we take through the camera using the app.


Gear61 (approved):
Oh my god, how did I mess this up?!? Awesome fix!!!

PR #173, opened 2020-09-06
Create Backup data manager

mohammedhemaid:
The video of the feature:
https://photos.app.goo.gl/ZgKHqSZjHdeAoAbV9

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataFragment.kt
+++ b/app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataFragment.kt
@@ -38,10 +38,10 @@ class BackupDataFragment : Fragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        backupDataClicked()
+        setUpBackupButtonOnClickListener()
         setBackupSubtitle()
         changeBackupFolder()
-        BackupDataManager.listener = backupDataListener
+        BackupDataManager.setListener(backupDataListener)
     }
 
     fun setBackupSubtitle() {
@@ -50,13 +50,11 @@ class BackupDataFragment : Fragment() {
         } else {
             val lastBackupUnixTime: Long = preferencesManager.lastBackupTime
             val lastBackupTime: String = TimeUtils.getLastBackupTime(lastBackupUnixTime)
-            binding.backupSubtitle.text = requireContext().getString(
-                    R.string.backup_subtitle_with_backup,
-                    lastBackupTime)
+            binding.backupSubtitle.text = getString(R.string.backup_subtitle_with_backup, lastBackupTime)
         }
     }
 
-    private fun backupDataClicked() {
+    private fun setUpBackupButtonOnClickListener() {
         binding.backupData.setOnClickListener {
             if (PermissionUtils.isPermissionGranted(Manifest.permission.WRITE_EXTERNAL_STORAGE, context)) {
                 backupData()
@@ -87,7 +85,7 @@ class BackupDataFragment : Fragment() {
         startActivityForResult(intent, WRITE_BACKUP_FILE_CODE)
     }
 
-   private fun changeBackupFolder() {
+    private fun changeBackupFolder() {
         binding.changeBackupFolder.setOnClickListener {
             if (PermissionUtils.isPermissionGranted(Manifest.permission.WRITE_EXTERNAL_STORAGE, context)) {
                 chooseBackupLocation()
@@ -113,9 +111,9 @@ class BackupDataFragment : Fragment() {
 
     override fun onActivityResult(requestCode: Int, resultCode: Int, resultData: Intent?) {
         if (requestCode == WRITE_BACKUP_FILE_CODE && resultCode == Activity.RESULT_OK && resultData != null) {
-            val context = context
+            val context = requireContext()
             val uri = resultData.data
-            if (uri == null || context == null) {
+            if (uri == null) {
                 backupDataListener.onBackupFailed()
             } else {
                 // Persist ability to read/write to this file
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
```diff
@@ -134,6 +134,38 @@ public void renameList(int listId, String newName) {
         close();
     }
 
+    public Map<ListDO, List<NameDO>> getAllListSet() {
+        Map<ListDO, List<NameDO>> allListMap = new HashMap<>();
+        ListDO listDO = new ListDO(-1, "");
```
Gear61: Why do we create a new listDO out here with these initial values? Why not do a local variable inside the for loop? This feels hard to read.

app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+        preferencesManager = PreferencesManager(context)
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        backupDataClicked()
+        setBackupSubtitle()
+        changeBackupFolder()
+        BackupDataManager.listener = backupDataListener
```
Gear61: Interesting, is BackupDataManager is singleton? I forgot how I built it for flashcards.

app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+            savedInstanceState: Bundle?
+    ): View? {
+        _binding = BackupDataBinding.inflate(inflater, container, false)
+        preferencesManager = PreferencesManager(context)
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        backupDataClicked()
```
Gear61:
```suggestion
        setUpBackupButtonOnClickListener()
```
The current name seems like something that would fire after the button has been clicked.

app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+        BackupDataManager.listener = backupDataListener
+    }
+
+    fun setBackupSubtitle() {
+        if (BackupDataManager.getBackupUri(context) == null) {
+            binding.backupSubtitle.setText(R.string.backup_data_explanation)
+        } else {
+            val lastBackupUnixTime: Long = preferencesManager.lastBackupTime
+            val lastBackupTime: String = TimeUtils.getLastBackupTime(lastBackupUnixTime)
+            binding.backupSubtitle.text = requireContext().getString(
```
Gear61:
```suggestion
            binding.backupSubtitle.text = getString(
```
We should just be able to do this.

app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+        override fun onBackupFailed() {
+            UIUtils.showLongToast(R.string.backup_failed, context)
+        }
+    }
+
+    override fun onActivityResult(requestCode: Int, resultCode: Int, resultData: Intent?) {
+        if (requestCode == WRITE_BACKUP_FILE_CODE && resultCode == Activity.RESULT_OK && resultData != null) {
+            val context = context
+            val uri = resultData.data
+            if (uri == null || context == null) {
```
Gear61: I don't think we should null check context here (feels like unnecessary error suppression). I think we should just do requireContext() in the block where we need it.

app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
```diff
@@ -134,6 +134,38 @@ public void renameList(int listId, String newName) {
...
+            if (listDO.getId() != listId){
+                listDO = new ListDO(listId, listName);
+                nameDOList = new ArrayList<>();
+            }
+            NameDO nameDO = new NameDO();
+            String name = cursor.getString(2);
+            nameDO.setName(name);
+            int nameAmount = cursor.getInt(3);
+            nameDO.setAmount(nameAmount);
+            nameDO.setPhotoUri(cursor.getString(4));
```
Gear61: I'm not sure if the photo URI will ever work when restoring from backup since it's app/device specific I think. If it doesn't work, we should just not back this up, similar to flashcards.

app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
```diff
@@ -134,6 +134,38 @@ public void renameList(int listId, String newName) {
         close();
     }
 
+    public Map<ListDO, List<NameDO>> getAllListSet() {
```
Gear61: Having a Map to represent a relationship between 2 classes seems hacky. I think we should just modify ListDO to have a list of NameDO.

Gear61 (changes requested):
Seems good overall! Also, I left feedback on the first PR around the UI, so if we can address that in this PR (or the next one), that would be great.

app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataFragment.kt
mohammedhemaid: hahaha yes, I used an object keyword instead of class, there will only ever be one instance of this class, and the instance which is created the first time it is accessed, in a thread-safe manner.

app/src/main/java/com/randomappsinc/studentpicker/backupandrestore/backup/BackupDataManager.kt
```diff
@@ -68,14 +72,10 @@ object BackupDataManager {
     }
 
     private fun alertListenerOfBackupComplete() {
-        if (listener != null) {
-            uiHandler.post { listener!!.onBackupComplete() }
-        }
+        uiHandler.post { listener?.onBackupComplete() }
```
Gear61, line 75: Oh yeah, we can do this.

Gear61 (approved):
Looks good!

PR #174, opened 2020-09-13
Backup and restore

mohammedhemaid:
Add ObjectSerializer, the data will be serialized as an encoded string in the backup file

### DIFF
--- a/app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
+++ b/app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
@@ -20,8 +20,10 @@ import com.randomappsinc.studentpicker.utils.JSONUtils;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public class DataSource {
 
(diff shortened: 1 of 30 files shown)


app/src/main/java/com/randomappsinc/studentpicker/database/DataSource.java
```diff
@@ -134,6 +136,40 @@ public void renameList(int listId, String newName) {
         close();
     }
 
+    public Set<ListDO> getAllNamesSet() {
+        Set<ListDO> allNamesSet = new HashSet<>();
```
Gear61, line 140: Hmm, I feel like this is sort of hard to read? If I were to do this, I would do a HashMap of list ID -> list. For every row returned by the query (name + list information), I would grab the list. If it's null, add to its list of names. Otherwise, instantiate a new list, attach the name to it, and put it in the map. At the end, you can return a List of ListDOs instead of a Set.

app/src/main/java/com/randomappsinc/studentpicker/settings/SettingsFragment.java
```diff
@@ -122,6 +124,16 @@ public void onItemClick(int position) {
             case 6:
                 intent = new Intent(Intent.ACTION_VIEW, Uri.parse(REPO_URL));
                 break;
+
```
Gear61, line 127: nit: Delete the extra newline?

Gear61 (approved):

PR #175, opened 2020-09-29
Backup and restore

mohammedhemaid:
I don't know if we should create a new branch for this PR, so I set the base as master 

The video for the changes:
https://photos.app.goo.gl/8dwepcamaskcGb89A


app/src/main/java/com/randomappsinc/studentpicker/utils/JSONUtils.kt
```diff
@@ -0,0 +1,74 @@
...
+            val listDOJson: JSONObject? = createNameListJson(listDO)
+            if (listDOJson != null) {
+                nameListArray.put(listDOJson)
+            }
+        }
+        return nameListArray.toString()
+    }
+
+    private fun createNameListJson(listDO: ListDO): JSONObject? {
+        try {
```
mohammedhemaid, line 54: The is no need to try and catch as JSON as the compiler compile the code without any issues and from what I remember Kotlin doesn't have checked exceptions

Gear61 (approved):
Looks good!

PR #178, opened 2020-11-21
Polish import from file activity progress dialog

mohammedhemaid:
Make use of common ProgressDialog in the import file screen


Gear61:
Makes sense to me!

Gear61 (approved):

PR #180, opened 2021-01-01
Add upsell dialog for backup and restore

mohammedhemaid:
A video of the feature:
https://photos.app.goo.gl/6EGuUC6YW9PqeJKB8

Set it once for mod 7


app/src/main/java/com/randomappsinc/studentpicker/ads/PremiumBackDoor.kt
```diff
@@ -0,0 +1,5 @@
+package com.randomappsinc.studentpicker.ads
+
+object PremiumBackDoor {
```
Gear61: Ah, this made it past the .gitignore because it's Kotlin. Can you update the .gitignore to ignore *PremiumBackDoor* and then remove this from the commit? If we push this, the backdoor will be public and people will be able to get premium for free if someone leaves it in a review or something.

app/build.gradle
```diff
@@ -6,8 +6,8 @@ android {
     compileSdkVersion 30
 
     def versionMajor = 4
-    def versionMinor = 3
-    def versionPatch = 10
+    def versionMinor = 4
+    def versionPatch = 0
```
Gear61: This should increase version patch, not version minor. This is how I see it:
1. Major = A complete overhaul of the app2
2. Minor = A new major feature (like groups) that is probably an entirely new screen
3. Patch = Tweaks

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -133,6 +141,17 @@ private void showPleaseRateDialog() {
                 .show();
     }
 
+    private void showBackupAndDialogFeature() {
```
Gear61: nit: We're just showing the dialog, not the feature itself.
```suggestion
    private void showBackupAndRestoreDialog() {
```

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -95,6 +98,11 @@ public void onSlide(@NonNull View bottomSheet, float slideOffset) {
             showPleaseRateDialog();
         }
 
+        if (preferencesManager.getNumAppOpens() % NUM_APP_OPENS_FOR_RESTORE_AND_BACKUP == 0 &&
```
Gear61: Even though this should never happen, we should chain this against the previous block with an else-if to be safe. Showing 2 dialogs on top of one another seems like a bad experience.

app/src/main/java/com/randomappsinc/studentpicker/home/BackupAndRestoreDialog.kt
```diff
@@ -0,0 +1,18 @@
+package com.randomappsinc.studentpicker.home
+
+import android.content.Context
+import com.afollestad.materialdialogs.MaterialDialog
+import com.randomappsinc.studentpicker.R
+
+class BackupAndRestoreDialog(context: Context, onPositiveClick: () -> Unit) {
```
Gear61: Chaining the creation of the class to showing the dialog seems confusing - I wouldn't guess that's what the class does based on the name. It also feels off to use a class since we don't need to maintain any semblance of state. Maybe we can move this to a util like DialogUtil and make the method something like .showBackupAndRestoreUpsell() that takes in a callback?

app/src/main/java/com/randomappsinc/studentpicker/utils/PreferencesManager.java
```diff
@@ -97,4 +98,10 @@ public boolean hasSeenPremiumTooltip() {
         prefs.edit().putBoolean(HAS_SEEN_PREMIUM_TOOLTIP, true).apply();
         return hasSeenTooltip;
     }
+
+    public boolean hasSeenBackupAndRestore() {
```
Gear61, line 106: Coupling the getting to the setting is a nice hack to reduce lines of code, but I feel like it's not readable and it breaks the contract set by the name of the method. An outside developer wouldn't guess that the API does this. Let's get away from this bad way of programming I introduced and introduce both a setter (which will set the value to true and can be called onBackupAndRestoreSeen() or something) and a getter. This will make some other code cleaner too (I will describe later).

app/src/main/res/values/strings.xml
```diff
@@ -34,6 +34,21 @@
     </string>
     <string name="will_rate">Sure, I\'ll help</string>
 
+    <string name="never_lost_your_data">Never Lose Your Data!</string>
```
Gear61: Instead of coupling the string name to the content of the string itself, it's better to describe how it's used IMHO, and you can see this in the other variable names (this is something we do at Instagram that I want to do more of). Here are the names we can use:

backup_and_restore_dialog_title
backup_and_restore_dialog_body
backup_and_restore_dialog_positive_button
backup_and_restore_dialog_negative_button

Gear61 (commented):
Since premium users have given us money, let's make sure we keep their experience top-notch and don't bother them with a useless dialog. Here are the 2 cases I am thinking of:
1. They already have this feature set up (important)
2. They have visited the page and don't want to opt in for whatever reason (less important)

Let's make sure that for these 2 cases, we don't show the dialog for them. We will need to extend our if statement. After we split the shared preferences API, we can call it when the page is visited to remember not to show the dialog.

Gear61 (changes requested):
Meant to request changes instead of just leave a comment. :(

app/src/main/java/com/randomappsinc/studentpicker/ads/PremiumBackDoor.kt
mohammedhemaid: you're right I should have to check it before pushing it.

app/build.gradle
mohammedhemaid: so even if the patch version is 10 we should increase it to 11 and not increase the minor version?

app/src/main/java/com/randomappsinc/studentpicker/home/BackupAndRestoreDialog.kt
mohammedhemaid: You're right, I spend some time thinking about this to do it in utils or a separate class then decided to put it in a class to match the same pattern we used before

app/src/main/java/com/randomappsinc/studentpicker/utils/PreferencesManager.java
mohammedhemaid: Awesome!

mohammedhemaid:
> Meant to request changes instead of just leave a comment. :(

I think you changed it to request changes, right?

app/src/main/java/com/randomappsinc/studentpicker/utils/PreferencesManager.java
```diff
@@ -99,9 +99,11 @@ public boolean hasSeenPremiumTooltip() {
         return hasSeenTooltip;
```
mohammedhemaid, line 99: I forgot to update this to make it more readable :(

app/build.gradle
Gear61: Yep!

app/src/main/java/com/randomappsinc/studentpicker/ads/PremiumBackDoor.kt
Gear61: We still need to remove this file from the branch. I think there's git rm --cached or something?

.gitignore
```diff
@@ -8,5 +8,6 @@
 app/release/
 Random-Name-Picker.iml
 *PremiumBackDoor.java
+*PremiumBackDoor.kt
```
Gear61: This is more scalable and less lines.
```suggestion
*PremiumBackDoor*
```

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -93,6 +95,10 @@ public void onSlide(@NonNull View bottomSheet, float slideOffset) {
         preferencesManager.increaseNumAppOpens();
         if (preferencesManager.getNumAppOpens() == NUM_APP_OPENS_FOR_RATING_ASK) {
             showPleaseRateDialog();
+        }else if (preferencesManager.getNumAppOpens() % NUM_APP_OPENS_FOR_RESTORE_AND_BACKUP == 0 &&
```
Gear61: super nit
```suggestion
        } else if (preferencesManager.getNumAppOpens() % NUM_APP_OPENS_FOR_RESTORE_AND_BACKUP == 0 &&
```

app/src/main/java/com/randomappsinc/studentpicker/utils/DialogUtil.kt
```diff
@@ -0,0 +1,20 @@
+package com.randomappsinc.studentpicker.utils
+
+import android.content.Context
+import com.afollestad.materialdialogs.MaterialDialog
+import com.randomappsinc.studentpicker.R
+
+object DialogUtil {
+
+    fun showBackupAndRestoreUpsell(context: Context, onPositiveClick: () -> Unit){
```
Gear61: super nit
```suggestion
    fun showBackupAndRestoreUpsell(context: Context, onPositiveClick: () -> Unit) {
```

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -133,6 +139,17 @@ private void showPleaseRateDialog() {
                 .show();
     }
 
+    private void showBackupAndRestoreDialog() {
+        DialogUtil.INSTANCE.showBackupAndRestoreUpsell(this, () -> {
```
Gear61, line 145: Just curious: What does INSTANCE do?

app/src/main/java/com/randomappsinc/studentpicker/utils/PreferencesManager.java
Gear61: No worries, we can worry about tooltip code later.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -93,6 +95,10 @@ public void onSlide(@NonNull View bottomSheet, float slideOffset) {
         preferencesManager.increaseNumAppOpens();
         if (preferencesManager.getNumAppOpens() == NUM_APP_OPENS_FOR_RATING_ASK) {
             showPleaseRateDialog();
+        }else if (preferencesManager.getNumAppOpens() % NUM_APP_OPENS_FOR_RESTORE_AND_BACKUP == 0 &&
+                !preferencesManager.hasSeenBackupAndRestore()){
```
Gear61: We should also check against whether the feature is already in use. I think we persist the backup file path to shared preferences, so we can check against that being non-null and non-empty.
```suggestion
                !preferencesManager.hasSeenBackupAndRestore()) {
```

app/src/main/java/com/randomappsinc/studentpicker/utils/PreferencesManager.java
```diff
@@ -97,4 +98,12 @@ public boolean hasSeenPremiumTooltip() {
         prefs.edit().putBoolean(HAS_SEEN_PREMIUM_TOOLTIP, true).apply();
         return hasSeenTooltip;
     }
+
+    public void onBackupAndRestoreSeen(boolean onSeen) {
```
Gear61, line 102: We need to call this API on the onCreate() of BackupAndRestoreActivity I think.

Gear61 (changes requested):
Thanks for iterating on the feedback! Hopefully just another round and we'll be good.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
mohammedhemaid: As DilaogUtil is an object file and we access it from Java so it deals with it as a singleton class that has one instance, once we convert HomeActivity to Kotlin we will not use INSTANCE.

The reason I used object for util is that there is no static method in Kotlin so, I made an object file that has one instance.

Gear61: Interesting, got it.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -93,6 +95,12 @@ public void onSlide(@NonNull View bottomSheet, float slideOffset) {
         preferencesManager.increaseNumAppOpens();
         if (preferencesManager.getNumAppOpens() == NUM_APP_OPENS_FOR_RATING_ASK) {
             showPleaseRateDialog();
+        } else if (preferencesManager.getNumAppOpens() % NUM_APP_OPENS_FOR_RESTORE_AND_BACKUP == 0 &&
+                       !preferencesManager.hasSeenBackupAndRestore() &&
+                            (preferencesManager.getBackupUri() == null
```
Gear61, line 100: TextUtils.isEmpty does both empty and null check in 1 method call.

app/src/main/java/com/randomappsinc/studentpicker/home/HomeActivity.java
```diff
@@ -93,6 +95,12 @@ public void onSlide(@NonNull View bottomSheet, float slideOffset) {
...
+                                    || preferencesManager.getBackupUri().isEmpty())) {
+            preferencesManager.onBackupAndRestoreSeen(true);
```
Gear61, line 102: Very small detail but this line should be in the method as they should be coupled together. You don't want someone calling the method without running this line of code.

Gear61 (approved):
A few small changes here and there but definitely good enough to merge and land. Awesome job!

PR #183, opened 2021-01-10
Add option to remove image from name

mohammedhemaid:
The video of the feature:

https://user-images.githubusercontent.com/25705658/104135473-8be7b880-5345-11eb-8869-d64f844b37a9.mp4


Gear61 (approved):
I didn't realize this change would be so easy, haha. We wrote some good abstractions. Nice job. Love the test plan!

